PR #11, opened 2023-01-18
initial docs site

salaxander:
Initial docs site containing content from the readme, as well as the walkthrough tutorial. Includes GitHub Action to deploy the site to GitHub Pages


sozercan:
a few general comments:

- please remove the placeholder svgs
- link to your branch with a working url
- remove duplicate content from readme and provide a link there

website/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+### Build
+
+```
+$ yarn build
+```
+
+This command generates static content into the `build` directory and can be served using any static contents hosting service.
+
+### Deployment
```
CodeMonkeyLeet, line 27: This part of the template is confusing to leave in place. We should discuss how we want to host this and update the instructions to match.

salaxander:
> A couple of questions:
> 
> * Would it be easier to manage a docs site as a separate repo under the org rather than a subdirectory of the code repo? Depending on the intended hosting model, it might be easier to control permissions or integrate with a hosting solution.
> * Related to the above, do we have a strategy for keeping the docs versioned with the releases?
> 

I think it makes it easy to host on a GitHub pages URL tied to the repo if we keep it as a sub directory. If we wanted to rely on other hosting solutions, it might make sense to go with a separate repo. As for versioned docs, Eraser is doing it so we can use that same mechanism. I just didn't include it in this initial PR

salaxander:
closing this while I iterate a bit more on my fork. Don't wanna be hitting this PR with a bunch more commits

CodeMonkeyLeet:
> I think it makes it easy to host on a GitHub pages URL tied to the repo if we keep it as a sub directory. If we wanted to rely on other hosting solutions, it might make sense to go with a separate repo. As for versioned docs, Eraser is doing it so we can use that same mechanism. I just didn't include it in this initial PR

@salaxander Sounds good ... assuming that the website is a subdirectory of this project, we might also want to update the automated release workflow (release.yml) so that it's tied to a versioned release of the docs as as separate work item.

PR #13, opened 2023-01-20
Doc site v1

salaxander:
Opening this again. This is an initial version to get the site up and running. I've cleared all but the basic info from the readme and pointed to the docs site. I will also aim to add versioning in a future PR.

You can see what this looks like running here: https://salaxander.github.io/copacetic

We will also need to add the redirect index.html file directly to the gh-pages branch. I will do that in another PR


sozercan:
looks like https://github.com/project-copacetic/copacetic/tree/gh-pages branch have stuff in it, can create an empty branch?

sozercan (approved):
LGTM!

PR #20, opened 2023-01-25
docs: add versioned docs

### LINKED ISSUE #15: [DOC] Add versioned docs
### What kind of documentation improvement is needed?

Other

### What is the change that is needed?

Update docs to be versioned similar to https://azure.github.io/eraser

salaxander:
Signed-off-by: Xander Grzywinski <xandergr@microsoft.com>

Add versioned docs starting with `v0.1.x`

Closes #15


CodeMonkeyLeet (commented):
Just questions for my understanding:

* Should we be using the `v?.?.x` versioning name? It looks like once Docusaurus spits out a versioned copy, those are expected to be frozen. If we only want to publish doc versions per minor update, then it feels like we should just do `v?.?`, which leaves open the possibility that we may publish `v?.?.?` revision to correct something particularly egregious in the minor version of the docs.

* Is the intention that the docusaurus versions are locked against the release versions? Just thinking about whether we should integrate the doc versioning with the release automation.

salaxander:
@sozercan how does Eraser do the release automation? I couldn't find anything in the repo that looked like it tied the docs to newly released versions via automation

sozercan:
@salaxander it is defined here: https://github.com/Azure/eraser/blob/main/.github/workflows/release-pr.yaml#L69

https://github.com/Azure/eraser/blob/main/Makefile#L301

website/versions.json
```diff
@@ -0,0 +1,3 @@
+[
+    "v0.1.x"
+]
```
CodeMonkeyLeet, line 3: Add missing newline.

Makefile
```diff
@@ -118,3 +119,16 @@ setup:
...
+################################################################################
+.PHONY: version-docs
+version-docs:
+	$(info $(INFOMARK) Creating versioned docs ...)
+	docker run --rm \
+		-v $(shell pwd)/website:/website \
+		-w /website \
+		$(IDFLAGS) \
+		node:${NODE_VERSION} \
+		sh -c "yarn install --frozen lockfile && yarn run docusaurus docs:version ${CLI_VERSION}"
```
CodeMonkeyLeet, line 134: Add missing newline ... eslint for go checks for this but not for general file types, and it's nice to have our source files play nice with POSIX tools that expect text files that end with newline.

website/versions.json
```diff
@@ -0,0 +1,3 @@
+[
+    "v0.1.x"
```
CodeMonkeyLeet, line 2: This should be `v0.1.0` given how the release automation flow works: `semrel.outputs.version` is inferred based on whether the branch contains non-breaking feature changes, so the next time we run it, the `CLI_VERSION` will be either `v0.2.0` or `v0.1.1` and the `?.?.x` format will be a weird one-off.

CodeMonkeyLeet (commented):
Looks generally good to me, thanks for adding the automation in - see comment about the v0.1.x version though.

CodeMonkeyLeet (approved):
LGTM, thanks!

PR #27, opened 2023-01-31
fix: update docusaurus dependencies

CodeMonkeyLeet:
docusaurus/core@2.2.0 has transitive dependencies on got@9.6.0 and trim@0.0.1 which have security vulnerabilities. Docusaurus 2.3.0 and canary do not have updates to these dependencies coming in the near future, so this change forces yarn resolution to the patched versions.

Signed-off-by: Simon Leet <simon.leet@microsoft.com>


sozercan (approved):
Nice find! LGTM

PR #45, opened 2023-02-22
ci: define semantic types and title only

sozercan:
Add semantic.yml to define accepted Semantic Commits types to enforce. See https://github.com/Ezard/semantic-prs.


CodeMonkeyLeet (approved):
Thanks for adding this!

PR #55, opened 2023-03-01
docs: add development-tips.md

CodeMonkeyLeet:
Add documentation for debugging and testing changes to `copa`.


sozercan:
@CodeMonkeyLeet can we move this to the website docs so it's more visible?

docs/development-tips.md
```diff
@@ -0,0 +1,254 @@
...
+`dive` provides a simple interface for walking the layers of the image and inspecting the files that were added or changed at each layer with your arrow keys.
+
+- `Tab` will toggle between navigating the layers and the files in the layer.
+- Filtering out unmodified files with `Ctrl+U` while in files view will effectively show you the file diff introduced by that layer.
+
+In particular, if you are adding or changing any of the patching functionality, the diff view of the files in the image can be useful to verify that the expected files have actually been written to the target image.
+
+### Extract files from the image to inspect them
+
+`dive` won't let you read the contents of the files in the image though; to do that, you can use the `docker cp` command to copy the files out of the image to a local folder. Note that `docker cp` only works with containers and not just container images, so you will need to create a container from the image and then copy the files out of the container:
```
sozercan, line 67: I have been using `crane export` to dump the filesystem. Should we mention this instead of `docker cp`?

You can also see diffs using
```
 diff \ 
    <(crane export foo:bar - | tar -tvf - | sort) \
    <(crane export foo:bar-patched - | tar -tvf - | sort)
 ```

CodeMonkeyLeet:
> @CodeMonkeyLeet can we move this to the website docs so it's more visible?

I'm open to it, I'll leave it to @salaxander's discretion. For other projects I've worked on, maintainer-facing docs tend to be in the repo and user-facing docs tend to be in the docs site.

docs/development-tips.md
CodeMonkeyLeet: I'll add it to the doc because `crane export` is a good way to explore the entire filesystem interactively, but I switched away from extracting the entire image in my workflow because there's usually only a couple of files I want to inspect, and the `dive` exploration view is really handy for the per layer diffing.

Also, neat `diff` command line, thanks for sharing that!

PR #56, opened 2023-03-02
docs: update governance .md files

CodeMonkeyLeet:
Update the governance policy documents to reflect the current state of the project and adoption of the GitHub Minimal Viable Governance (MVG) policies. See https://github.com/github/MVG for more information.


GOVERNANCE.md
```diff
@@ -0,0 +1,45 @@
...
+
+Information disclosed in connection with any Project activity, including but not limited to meetings, contributions, and submissions, is not confidential, regardless of any markings or statements to the contrary.
+
+## 5. Trademarks.
+
+Any names, trademarks, logos, or goodwill developed by and associated with the Project (the "Marks") are controlled by the Organization. Maintainers may only use these Marks in accordance with the Organization's trademark policy. If a Maintainer resigns or is removed, any rights the Maintainer may have in the Marks revert to the Organization.
+
+## 6. Amendments.
+
+Amendments to this governance policy may be made by affirmative vote of 2/3 of all Maintainers, with approval by the Organization's Steering Committee.
```
CodeMonkeyLeet, line 41: @sozercan @salaxander This is the boilerplate default and we should definitely make sure that we agree with this before signing off on this.

salaxander (approved):
Looks good to me. I think the only question would be that the governance doc refers to a steering committee. Do you think it's worth documenting that in this case the steering committee would be the maintainers?

CodeMonkeyLeet:
> Looks good to me. I think the only question would be that the governance doc refers to a steering committee. Do you think it's worth documenting that in this case the steering committee would be the maintainers?

@salaxander Good idea, I've added a clarifying preamble to the governance doc, let me know what you think.

sozercan:
@salaxander is this still valid? if not, can we close it?

salaxander:
> @salaxander is this still valid? if not, can we close it? or merge?

Yep no longer needed with CNCF governance. Closing.

PR #60, opened 2023-03-09
fix: add apk add

sozercan:
Add apk packages before upgrading. This works around the issue where packages require other packages in the dependency tree to be updated.


CodeMonkeyLeet (approved):
LGTM! Interesting find that just `apk add` fixes up the dependency tree for the package independent of versioning.

PR #79, opened 2023-03-22
feat: add http proxy support when executing package managers

### LINKED ISSUE #80: [REQ]  Add http proxy support when executing package managers

I would like to add http proxy support when executing package manager updates, so we can use this tool in environments using corporate proxies.

derdanne:
Add http proxy support when executing package managers.

Closes #80

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -9,6 +9,8 @@ import (
 	"io/fs"
 	"os"
 	"path/filepath"
+
+	"github.com/moby/buildkit/client/llb"
 )
 
 func EnsurePath(path string, perm fs.FileMode) (bool, error) {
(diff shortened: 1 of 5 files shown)


pkg/utils/utils.go
```diff
@@ -36,3 +38,22 @@ func IsNonEmptyFile(dir, file string) bool {
 	}
 	return !info.IsDir() && info.Size() > 0
 }
+
```
sozercan, line 41: thanks for the PR! Can we add unit tests for these please?

derdanne: Done!

sozercan (approved):
Thanks! LGTM

sozercan:
One of the tests are failing, but I don't think this is related to this. Investigating.

PR #119, opened 2023-04-26
ci: rename release-pr.yaml to release-docs.yaml file

### LINKED ISSUE #113: [REQ] update release-pr to release-docs

Other

https://github.com/project-copacetic/copacetic/blob/main/.github/workflows/release-pr.yml

consider updating release-pr to something like release-docs since workflow is different in this repo

reetasingh:
rename release-pr.yaml to release-docs.yaml file as it updates docs vs updating actual binary

Closes #113


sozercan:
@reetasingh thanks for the PR! steps also needs updating so we basically run "make version-docs ..." in the release that was cut.

reetasingh:
> #113

@sozercan do we want 

> @reetasingh thanks for the PR! steps also needs updating so we basically run "make version-docs ..." in the release that was cut. (current logic is using vNext tag)


@sozercan  I'm unsure if I have a complete understanding of the requirement. From the analysis of the workflow, it appears that when the workflow is manually triggered (indicated by github.event_name being equal to 'workflow_dispatch'), the NEW_VERSION value is determined based on the release_version input provided. However, when a new version tag is pushed (indicated by github.event_name being equal to 'push'), the current logic uses the vNext tag for NEW_VERSION. 

Do we want for the second case (when a new version tag is pushed), to not increment the minor version but rather to use it as it is?





Regenerate response

PR #237, opened 2023-08-08
feat: push new copa-action image on release

ashnamehrotra:
To simplify action releasing, instead of releasing from copa-action on every copa release, we can push the new copa-action image on each release of copa (https://github.com/orgs/project-copacetic/packages/container/package/copa-action). The copa-action releases will not need to be created often, since it will stay the same and just requires an updated copa-action image (with the new copa version) to pull from the new releases.

Closes #<_issue_ID_>


.github/workflows/release.yml
```diff
@@ -36,3 +36,18 @@ jobs:
...
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Checkout copa-action repository
+        uses: actions/checkout@v2
+        with:
+          repository: project-copacetic/copa-action
+          ref: main
+
+      - name: Set up Docker
+        uses: docker/setup-buildx-action@ecf95283f03858871ff00b787d79c419715afc34
```
sozercan, line 47: do we need this if we are not using buildx?

sozercan: or we can use buildx below and build and push directly

.github/workflows/release.yml
```diff
@@ -36,3 +36,17 @@ jobs:
           args: release --clean --config .goreleaser.yml
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Checkout copa-action repository
+        uses: actions/checkout@v2
```
sozercan, line 41: nit: please add a tag and comment with the corresponding version

```suggestion
        uses: actions/checkout@12345 # vX.Y.Z
```

.github/workflows/release.yml
```diff
@@ -36,3 +36,17 @@ jobs:
...
+        with:
+          repository: project-copacetic/copa-action
+          ref: main
+
+      - name: Set up Docker
+        uses: docker/setup-buildx-action@ecf95283f03858871ff00b787d79c419715afc34
```
sozercan, line 47: comment with version please

sozercan (approved):
LGTM pending tests

PR #261, opened 2023-08-21
feat: add modular scanners

### LINKED ISSUE #59: [REQ] modular scanners

New feature

It would be good to have scanners to be modular so they don't have to live in the core copacetic repository. The intention is for others to contribute without each scanner being part of the core copa.

perhaps a model like https://github.com/deislabs/ratify/blob/main/pkg/verifier/plugin/skel/skel.go https://github.com/deislabs/ratify/blob/main/plugins/verifier/licensechecker/licensechecker.go (example) which defines inputs and outputs might work

anubhav06:
Added support for modular scanners. 

Closes #59

## Overview
- Trivy remains the default scanner.
- If someone wants to use vulnerability data from other scanners, they can utilize plugins.
- Plugins can be written and used for using vulnerability data of other scanners. The plugin is an executable binary and must be available on the $PATH for copa to execute.
- The command which can be used to invoke a plugin is as follows:
```
copa patch --scanner copa-grype -i docker.io/library/nginx:1.21.6 -r nginx.1.21.6.json -t nginx.1.21.6-patched
```
- I've written the Grype plugin, which can be used for patching grype produced results: 
**copa-grype plugin:** https://github.com/anubhav06/copa-grype

## How does it Work?
- The plugin takes the image vuln report path as an argument. 
- The plugin then parses the report and outputs in a format supported by copa ([Update Manifest struct](https://github.com/project-copacetic/copacetic/blob/main/pkg/types/types.go#L15C6-L15C21))
- Copa executes the plugin specified by the user, and captures the output of the plugin, thereby having access to the required updates. It then simply passes this info further to the copa patching process.

## How to Test?
- Build the [copa-grype plugin](https://github.com/anubhav06/copa-grype) to produce a binary called `copa-grype`
- Then execute copa(this version of PR) with the following command:
```
copa patch --scanner /path/to/copa-grype -i <image> -r <grype-results>
```

After adding the `copa-grype` binary to PATH:
267555333-1217c222-9359-458d-a36a-b8a7fd270000


anubhav06:
This is how it looks when I add the copa-grype binary to my PATH and then run it:

image

anubhav06:
The owner in https://github.com/anubhav06/copa-grype will need to be changed from `anubhav06` to `project-copacetic`

sozercan:
> The owner in [anubhav06/copa-grype](https://github.com/anubhav06/copa-grype) will need to be changed from `anubhav06` to `project-copacetic`

This is the motivation of the modular scanners so we don't have to. However, we should add a doc to document existing integrations (example: https://open-policy-agent.github.io/gatekeeper/website/docs/externaldata#providers-maintained-by-the-community)

pkg/report/report.go
```diff
@@ -21,7 +23,36 @@ type ScanReportParser interface {
 	Parse(string) (*types.UpdateManifest, error)
 }
 
-func TryParseScanReport(file string) (*types.UpdateManifest, error) {
+func TryParseScanReport(file, scanner string) (*types.UpdateManifest, error) {
```
sozercan, line 26: can you add a test provider (can use existing trivy code) and make sure to add tests

pkg/patch/cmd.go
```diff
@@ -52,6 +55,7 @@ func NewPatchCmd() *cobra.Command {
 	flags.StringVarP(&ua.workingFolder, "working-folder", "w", "", "Working folder, defaults to system temp folder")
 	flags.StringVarP(&ua.buildkitAddr, "addr", "a", "", "Address of buildkitd service, defaults to local docker daemon with fallback to "+buildkit.DefaultAddr)
 	flags.DurationVar(&ua.timeout, "timeout", 5*time.Minute, "Timeout for the operation, defaults to '5m'")
+	flags.StringVarP(&ua.scanner, "scanner", "s", "", "Scanner used to generate the report, defaults to 'trivy'")
```
sozercan, line 58: if it defaults to trivy, we should add a default

pkg/report/report.go
```diff
@@ -21,7 +23,36 @@ type ScanReportParser interface {
...
+
+	if scanner == "" {
```
sozercan, line 28: you'll need to handle `trivy` case

anubhav06:
> @anubhav06 This is the motivation of the modular scanners so we don't have to (unless we want to officially support an integration). However, we should add a doc to document existing integrations (example: https://open-policy-agent.github.io/gatekeeper/website/docs/externaldata#providers-maintained-by-the-community)

Oh ok understood. That's interesting. I'll add the docs as well, once this PR is complete and merged.

anubhav06:
> @anubhav06 looks like this is not working for me

This is because in the new version of grype (v0.69.1) which you are using, it seems they have removed "grype" from `descriptor.name`.
I had tested this with v0.64.2, in that version `descriptor.name` had the value "grype". 

It's strange as to why did they remove this in the new version, I'll look into this and will get back with a solution 👍

pkg/report/report.go
anubhav06: there is also a `QualysParser`, do we want it to remain as default like trivy (like it is now), or it should be removed and added as a plugin?

pkg/report/report.go
anubhav06: just to confirm, I should add a provider file like https://github.com/open-policy-agent/gatekeeper/blob/master/test/externaldata/dummy-provider/provider.go right? sorry if this is a silly question as I'm new to test providers

pkg/report/report.go
sozercan: I am okay to remove to qualys as we don't have access to it anymore

pkg/report/report.go
sozercan: no worries! we don't need the same or similar code (just the concept). if we can turn trivy into a test provider and without duplicating code, i think that would be ideal so we can make sure to test the provider call flow.

anubhav06: can you please check my new commits regarding tests, is it what is expected?

anubhav06:
For testing:
1. Download the newer version of copa-grype: https://github.com/anubhav06/copa-grype
2. Add "grype" to `descriptor.name` field of your grype output. (until the grype bug is fixed)

pkg/report/provider.go
```diff
@@ -0,0 +1,26 @@
+package report
```
sozercan, line 1: Is this file used?

pkg/report/report.go
sozercan: yes! this is great for now. we can do a follow up to integrate https://github.com/project-copacetic/scanner-parser-template/ to test success cases into e2e

as a follow up, we'll need:
- go install https://github.com/project-copacetic/scanner-parser-template in github actions e2e
- run with a fake report
- make sure copa can call and receive results from the plugin

pkg/report/provider.go
anubhav06: No. I was not able to clearly understand the dummy-provider reference repo, so I was not sure if I should add this function or not.

pkg/report/report.go
```diff
@@ -21,7 +23,36 @@ type ScanReportParser interface {
...
+	scannerCommand := exec.Command(cmd, file)
+	// Capture the output
+	scannerOutput, err := scannerCommand.CombinedOutput()
+
+	if err != nil {
+		return nil, fmt.Errorf("error running scanner %s: %w", scanner, err)
+	}
+	// Convert the output to a UpdateManifest struct
+	var um types.UpdateManifest
+	if err := json.Unmarshal(scannerOutput, &um); err != nil {
```
sozercan, line 47: from the specs, we need to version our api so our plugins can communicate with copa with an API

here are the steps:
- for types, let's split this into 2 packages (`v1alpha1` and `unversioned`). unversioned is the internal api where we would convert everything to. Directory structure would look something like:
```shell
types
├── unversioned
│   └── types.go
└── v1alpha1
    ├── convert.go
    └── types.go
```

- copy `types.go` into each (since they are exactly same for now) and create a const with an apiVersion (this is what providers can use, example: https://github.com/project-copacetic/scanner-parser-template/blob/apiVersion/main.go#L41C23-L41C23)

```go
package v1alpha1
const APIVersion = "v1alpha1"
...
```

and for unversioned

```go
package unversioned
const APIVersion = "unversioned"
...
```
- make sure to update all internal types references to use `unversioned`.  Idea is that plugin communicate via API then copa can use it's own internal API.

- instead of just unmarshalling here, first we want to check what `apiVersion` we got from the plugin so we can convert later on. We can unmarshal the output to `map[string]interface{}{}` 

- then create a function to take this output and we can convert to unversioned/internal api. something like:
```go
func convertToUnversionedAPI(scannerOutput []byte, m map[string]interface{}) (*unversioned.UpdateManifest, error) {
	switch v := m["apiVersion"].(type) {
	case string:
		if v == "v1alpha1" {
			um, err := v1alpha1.Convert_v1alpha1_UpdateManifest_To_unversioned_UpdateManifest(scannerOutput)
...
<for other versions, we would error out as unsupported>
```

- We can define conversion func in the `types/unversioned/convert.go`, which is basically just unmarshalling to the `unversioned` since they are same at this time.

let me know if you have any questions. feel free to reach out over slack.

sozercan (changes requested):
see comment

pkg/report/provider.go
sozercan: let's remove this  if it's not used

anubhav06: removed

anubhav06:
sorry, work still in progress, kindly do not review yet

pkg/report/report.go
anubhav06: we will need to release a version of the scanner-parser-template if we want to install it for testing, right?
For example: In the file `.github/workflows/scripts/download-tooling.sh` we can install it using something like this:
```bash
curl -sfL https://github.com/project-copacetic/scanner-parser-template/releases/download/v${SCANNER_PARSER_TEMPLATE_VERSION}/scanner-parser-template_${SCANNER_PARSER_TEMPLATE_VERSION}_Linux-64bit.tar.gz -o scanner-parser-template.tar.gz
sudo tar -zxvf scanner-parser-template.tar.gz -C /usr/local/bin/
rm scanner-parser-template.tar.gz

**Comment by sozercan** [comment]
@anubhav06 sorry for delay, we had a few CI blockers! i pushed a minor update to your PR so we can get this merged asap, hope that's okay

**Review by sozercan — APPROVED** [review]
LGTM pending tests

**Comment by sozercan** [comment]
@anubhav06 merged, thank you!

File: pkg/report/report.go (hunk 1 above)
Comment (anubhav06) [reply pkg/report/report.go]: @sozercan in order to go install, we will need a release of the scanner-plugin-template, right?
We can also install using something like:
```bash
curl -sfL https://github.com/project-copacetic/scanner-plugin-template/releases/download/v${SCANNER_PLUGIN_TEMPLATE_VERSION}/scanner-plugin-template_${SCANNER_PLUGIN_TEMPLATE_VERSION}_Linux-64bit.tar.gz -o scanner-plugin-template.tar.gz
sudo tar -zxvf scanner-plugin-template.tar.gz -C /usr/local/bin/
rm scanner-plugin-template.tar.gz
```
similar to how [other tools are installed](https://github.com/project-copacetic/copacetic/blob/main/.github/workflows/scripts/download-tooling.sh) in the copa CI, however even that would require a release of the fake-plugin.

PR #270, opened 2023-08-24
draft 2: modular plugins

anubhav06:
Another possible (& better) approach for modular scanners is to use [go plugins](https://pkg.go.dev/plugin).
This PR demonstrates the use of go plugins for scanner support, instead of using `os/exec` pkg

Advantages:
- better than using `os/exec`. The data exchange process in  https://github.com/project-copacetic/copacetic/pull/261 is complicated. We don't have to marshall and unmarshall the output like we do there.

Disadvantages:
- Go plugins are only supported on Linux & MacOS. However, buildkitd already has a dependency that it works only on Linux, so this might not really be a disadvantage.
- `CGO_ENABLED` should be set to True. This leads to a slight increase in the size of produced copa binary

## How to Test?
1. Build copa: `make build`
2. Build plugin `make plugin`
```shell
copa patch --scanner /path/to/plugin -i <image> -r <grype-results>
```


salaxander:
I think this is an interesting approach, but there are some additional disadvantages to consider. Mainly that plugins are no longer language independent. With the exec binary approach, scanner plugins wouldn't be required to be written in Go. That may not be a problem at the end of the day, but it is a consideration. CC @sozercan

plugins/grype/copa-grype.go
```diff
@@ -0,0 +1,62 @@
+package main
```
sozercan, line 1: this would require all plugins to be in-tree, is that right?

anubhav06: right, like the following:

plugins
├── grype
│   └── copa-grype.go
├── plugin2
│   └── copa-plugin2.go
└── ...

anubhav06: Nope, not necessary for it to be in-tree. We can have a separate repo for each plugin just like in os exec approach.
For example: [https://github.com/anubhav06/copa-grype](https://github.com/anubhav06/copa-grype). This repo contains code for building a binary which can be run using the os exec approach. Instead, we can have the code for building the `.so` binary which can be run using the go plugins (this) approach.

However, now I'm thinking that we cannot add a `.so` file to our $PATH, right? Unlike os exec approach where the binary can be added to the path and can be automatically invoked; here we will have to specify the path of the `.so` file when running copa (--scanner <path_to_so-plugin_file> ). So this might not be a good approach

salaxander: If we specify a directory where the plugin files need to be, that could work. Like `~/.copa/plugins`.

anubhav06:
We can now just specify the plugin name, and copa will automatically look for it in the path `~/.copa/plugins/`. 
When downloading the plugin, we'll have to make sure that the plugin is installed in this PATH. (we can specify that in the shell script of downloading the plugin)

This is how it looks:
image

anubhav06:
Another downside might be that we might have to specify how to build plugins. 
For example using `go build -buildmode=plugin` or specify a Makefile like [this](https://github.com/project-copacetic/copacetic/pull/270/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R142-R154).

anubhav06:
Closed as we're going with the 1st approach, as discussed in the meet :)

PR #284, opened 2023-09-06
feat:  mTLS-over-TCP buildkit

RealHarshThakur:
Adds flags to enable Copa to connect to a remote Buildkit listening on TCP address secured via mTLS. 
This is very similar to creating a remote buildx driver as described [here](https://docs.docker.com/build/drivers/remote/#example-remote-buildkit-in-docker-container). 

Closes https://github.com/project-copacetic/copacetic/issues/285


pkg/buildkit/drivers.go
```diff
@@ -27,17 +27,40 @@ var (
...
+	client, err := client.New(ctx, bkOpts.Addr, opts...)
 	if err != nil {
 		return nil, err
 	}
 	return client, nil
 }
 
+func getCredentialOptions(bkOpts Opts) []client.ClientOpt {
+	opts := []client.ClientOpt{}
+	// addr := bkOpts.Addr
```
sozercan, line 44: looks like leftover comment?

RealHarshThakur: sorry, missed it. Removed it now :)

RealHarshThakur:
Guessing the test failures are unrelated to this change? As I see them here as well: https://github.com/project-copacetic/copacetic/pull/287 

Please do let me know if you folks think there is a test you'd like to see. I've followed the same pattern that buildctl CLI has to set up the client.

salaxander:
> Guessing the test failures are unrelated to this change? As I see them here as well: #287
> 
> Please do let me know if you folks think there is a test you'd like to see. I've followed the same pattern that buildctl CLI has to set up the client.

Yes it's happening across all PRs at the moment, working on addressing it

salaxander:
This is looking good to me - @sozercan do you have a chance to review?

pkg/patch/cmd.go
```diff
@@ -21,15 +21,18 @@ import (
...
-	buildkitAddr  string
-	timeout       time.Duration
-	ignoreError   bool
-	format        string
-	output        string
+	appImage               string
+	reportFile             string
+	patchedTag             string
+	workingFolder          string
+	buildkitAddr           string
```
cpuguy83, line 28: Instead of explicitly defining these addr/cert args here, can use `buildkit.Opts` and set the flags to use `&bkopts.<Field>`

RealHarshThakur: done :D

cpuguy83 (approved):

sozercan (approved):
Thank you! LGTM

Issue #285, opened 2023-09-06
[REQ] Support for Copa to connect to remote buildkit


RealHarshThakur:
Really cool project :)

### What kind of request is this?

Improvement of existing experience

### What is your request or suggestion?

Hey folks, I'd like to use Copa using a remote buildkit daemon, if people have builds and patching done over the same buildkit instance, they'll benefit from the cache.  

I've submitted a PR to support it: https://github.com/project-copacetic/copacetic/pull/284

RealHarshThakur:
On a separate note, I also realised that even with remote buildkit in place, this will still require docker daemon to run due to this bit [here]( https://github.com/project-copacetic/copacetic/blob/main/pkg/patch/patch.go#L141). 

I wonder if we can instead have a "push to registry" step there optionally. I'm happy to create a different issue and a PR to support it too, wanted to check if you folks have any opinions around it.

PR #294, opened 2023-09-07
feat: add push flag to support pushing to registry directly

### LINKED ISSUE #293: [REQ] Support pushing to registry directly

It would be nice to push to the image registry directly. 
Like : 
```
copa patch -i docker.io/library/alpine:3.16.4 -r patch.json -p ttl.sh/alpinepatched:1h
```

One side-effect of this would be that copa wouldn't rely on Docker anymore if there is a remote buildkit setup as I mentioned here. https://github.com/project-copacetic/copacetic/issues/285#issuecomment-1708239927

RealHarshThakur:
This PR enables direct push to registry from buildkit:
```
copa patch -i docker.io/library/alpine:3.16.4 -r patch.json -p ttl.sh/alpinepatched:1h
```

One side-effect of this would be that copa wouldn't rely on Docker anymore if there is a remote buildkit setup as I mentioned here. https://github.com/project-copacetic/copacetic/issues/285#issuecomment-1708239927

Closes #293

### DIFF
--- a/pkg/patch/patch.go
+++ b/pkg/patch/patch.go
@@ -133,8 +133,14 @@ func patchWithContext(ctx context.Context, image, reportFile, patchedTag, workin
 		return err
 	}
 
-	if err = buildkit.SolveToDocker(ctx, config.Client, patchedImageState, config.ConfigData, patchedImageName); err != nil {
-		return err
+	if pushDest != "" {
+		if err = buildkit.SolveToRegistry(ctx, config.Client, patchedImageState, config.ConfigData, pushDest); err != nil {
+			return err
+		}
+	} else {
+		if err = buildkit.SolveToDocker(ctx, config.Client, patchedImageState, config.ConfigData, patchedImageName); err != nil {
+			return err
+		}
 	}
 
 	// create a new manifest with the successfully patched packages
--- a/pkg/patch/patch.go
+++ b/pkg/patch/patch.go
@@ -5,12 +5,13 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"strings"
 	"time"
 
+	ref "github.com/distribution/reference"
 	log "github.com/sirupsen/logrus"
 	"golang.org/x/exp/slices"
 
-	"github.com/distribution/reference"
 	"github.com/project-copacetic/copacetic/pkg/buildkit"
 	"github.com/project-copacetic/copacetic/pkg/pkgmgr"
 	"github.com/project-copacetic/copacetic/pkg/report"
@@ -24,13 +25,13 @@ const (
 )
 
 // Patch command applies package updates to an OCI image given a vulnerability report.
-func Patch(ctx context.Context, timeout time.Duration, image, reportFile, patchedTag, workingFolder, scanner, format, output string, ignoreError bool, bkOpts buildkit.Opts) error {
+func Patch(ctx context.Context, timeout time.Duration, image, reportFile, patchedTag, workingFolder, scanner, format, output string, ignoreError, push bool, bkOpts buildkit.Opts) error {
 	timeoutCtx, cancel := context.WithTimeout(ctx, timeout)
 	defer cancel()
 
 	ch := make(chan error)
 	go func() {
-		ch <- patchWithContext(timeoutCtx, image, reportFile, patchedTag, workingFolder, scanner, format, output, ignoreError, bkOpts)
+		ch <- patchWithContext(timeoutCtx, image, reportFile, patchedTag, workingFolder, scanner, format, output, ignoreError, push, bkOpts)
 	}()
 
 	select {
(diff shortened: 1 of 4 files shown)


sozercan:
I am wondering if we should align with buildx CLI here. Ideally, we don't want to break existing users so what I was thinking was to use `--tag` to accept `registy/repo:tag` and `--push` can be a bool. to remain backwards compat, we can parse if `--tag` includes `:` and `/`. If it includes both `:` and `/`, retag the entire image. If it doesn't include both `:` and `/`, only replace the tag.

`-i foo.io/repo/image:tag -t -patched` -> foo.io/repo/image:tag-patched (what we have today)
`-i foo.io/repo/image:tag -t bar.io/repo/image:tag` ->  bar.io/repo/image:tag
`-i foo.io/repo/image:tag -t bar.io/repo/image` -> invalid (must provide tag)
`-i foo.io/repo/image:tag -t repo/image:tag` -> invalid (must provide registry)

@jeremyrickard @cpuguy83 wdyt?

jeremyrickard:
> I am wondering if we should align with buildx CLI here. Ideally, we don't want to break existing users so what I was thinking was to use `--tag` to accept `registy/repo:tag` and `--push` can be a bool. to remain backwards compat, we can parse if `--tag` includes `:` and `/`. If it includes both `:` and `/`, retag the entire image. If it doesn't include both `:` and `/`, only replace the tag.
> 
> `-i foo.io/repo/image:tag -t -patched` -> foo.io/repo/image:tag-patched (what we have today) `-i foo.io/repo/image:tag -t bar.io/repo/image:tag` -> bar.io/repo/image:tag `-i foo.io/repo/image:tag -t bar.io/repo/image` -> invalid (must provide tag) `-i foo.io/repo/image:tag -t repo/image:tag` -> invalid (must provide registry)
> 
> @jeremyrickard @cpuguy83 wdyt?

I like the `-i foo.io/repo/image:tag -t bar.io/repo/image:tag` syntax. 

I also think that adding a --push flag would be useful, especially if it didn't require docker locally.  Not the same scenario, but the experience with`ko` and pushing images w/o docker is super useful.

RealHarshThakur:
cool, happy to go with what you folks suggested. 

Although, I am not sure about this bit:
```
-i foo.io/repo/image:tag -t repo/image:tag -> invalid (must provide registry)
```

This might still be what the user wants, partially due to how popular dockerhub is: `username/image:tag` is valid over there. Also, other such registries like `ttl.sh` also don't mind that format.

RealHarshThakur:
I've pushed a commit to get most of what you folks suggested. Added couple test cases that shows we'll be backwards-compatible and a new case for specifying a different registry. I've tried to stay unopinionated about destination registry conventions as copa will throw an error during push if buildkit fails to push anyway. Happy to take your suggestions tho

RealHarshThakur:
I think setting up the mock buildkit API  in unit tests failed, is it acting flaky?

sozercan:
> This might still be what the user wants, partially due to how popular dockerhub is: username/image:tag is valid over there. Also, other such registries like ttl.sh also don't mind that format.

this is being consistent with `-i` since you must provide `docker.io` there

RealHarshThakur:
> this is being consistent with -i since you must provide docker.io there

cool, I've added the registry check  @sozercan

RealHarshThakur:
@sozercan @jeremyrickard don't mean to badger, is there anything else you'd like me to change on this PR?

salaxander:
@RealHarshThakur sorry for the delay - @sozercan has been on vacation for the past couple days. I'll see if I can get you a review asap :)

pkg/patch/patch_test.go
```diff
@@ -49,3 +49,58 @@ func TestRemoveIfNotDebug(t *testing.T) {
...
+			name:       "tag passed with value",
+			image:      "docker.io/library/nginx:1.21.3",
+			patchedTag: "custom",
+			want:       "docker.io/library/nginx:custom",
+			wantErr:    false,
+		},
+		{
+			name:       "tag passed but without registry or repo",
+			image:      "docker.io/library/nginx:1.21.3",
+			patchedTag: "my.registry/nginx:1.21.3-patched",
```
sozercan, line 79: this should be valid (something like `myregistry.azurecr.io/nginx:1.21.3-patched`)

pkg/patch/patch.go
```diff
@@ -160,3 +147,45 @@ func patchWithContext(ctx context.Context, image, reportFile, patchedTag, workin
...
+		} else {
+			patchedTag = fmt.Sprintf("%s-%s", tag, defaultPatchedTagSuffix)
+		}
+	}
+
+	slashCount := strings.Count(patchedTag, "/")
+	if slashCount > 0 {
+		if slashCount < 2 {
+			err := fmt.Errorf("invalid tag %s, must be in the form <registry>/<image>:<tag>", patchedTag)
+			log.Error(err)
```
sozercan, line 181: i think this is duplicate, since return is logged already?

sozercan (changes requested):
Sorry about the delay! Please see comments

sozercan:
ping @RealHarshThakur. we are getting close to cutting v0.5.0 release if you want this to be included

RealHarshThakur:
I'll try but most likely,  I'll be able to pick it up by end of the month

RealHarshThakur:
@sozercan  sorry for the delay, I've made the changes you've requested.

sozercan:
Closing due to inactivity. Please feel free to re-open if there's ongoing work. Thanks!

PR #295, opened 2023-09-07
fix: handle mariner zstd compression

sozercan:

handle mariner zstd compression as `busybox rpm` does not support zstd compression
> Switch rpm package building to use to zstd compression level 7
https://github.com/microsoft/CBL-Mariner/releases/tag/2.0.20230805-2.0

this should fix the broken e2e

Closes #<_issue_ID_>


salaxander (approved):
🎉  Nice to see DCO is working again too 😆

PR #364, opened 2023-10-10
fix: ignore cache for "apt update"

cpuguy83:
"apt update" should never be cached.
This also shouldn't break caching for anything run after it unless the
rootfs changed (ie. apt cache was actually updated).


sozercan (approved):
Thank you! LGTM

PR #531, opened 2024-03-15
feat: ignore unpatchable packages and add GetUniqueLatestUpdates test

### LINKED ISSUE #499: [REQ] copa should fast fail if input has no fixable vulns

today, copa tries to patch an image given an input with no fixable vulns. this is a no op action.

for trivy, copa should check if there are any fixable vulns in the report before attempting to patch. if not, copa should fast fail and give feedback to the user that there are no fixable vulns.

for custom scanner, we should add this to docs as a best practice.

sozercan:

Closes #499


ashnamehrotra (approved):
LGTM

PR #532, opened 2024-03-16
chore: update buildkit to 1.13.1

sozercan:

- updates buildkit to 0.13.1
- debug uses plain mode

Closes #<_issue_ID_>


pkg/patch/patch.go
```diff
@@ -231,13 +231,24 @@ func patchWithContext(ctx context.Context, ch chan error, image, reportFile, pat
...
+	if cn, err := console.ConsoleFromFile(os.Stderr); err == nil {
+		c = cn
+	}
+
+	mode := progressui.AutoMode
+	if log.GetLevel() >= log.DebugLevel {
+		mode = progressui.PlainMode
+	}
+
+	display, err := progressui.NewDisplay(c, mode, progressui.WithPhase(c.Name()))
```
cpuguy83, line 244: Not sure `c.Name()` is the right thing here.
Can probably leave it without a phase set.

cpuguy83 (changes requested):

pkg/patch/patch.go
```diff
@@ -231,13 +231,22 @@ func patchWithContext(ctx context.Context, ch chan error, image, reportFile, pat
...
+		c = cn
+	}
+	mode := progressui.AutoMode
+	if log.GetLevel() >= log.DebugLevel {
+		mode = progressui.PlainMode
+	}
+	display, err := progressui.NewDisplay(c, mode)
+	if err != nil {
+		return err
+	}
```
cpuguy83, line 245:
```suggestion
	out := os.Stderr
	if cf, _ := console.ConsoleFromFile(os.Stderr); cf != nil {
		out = cf
	}
	mode := progressui.AutoMode
	if log.GetLevel() >= log.DebugLevel {
		mode = progressui.PlainMode
	}
	display, err := progressui.NewDisplay(out, mode)
	if err != nil {
		return err
	}
```

cpuguy83 (changes requested):

pkg/patch/patch.go
sozercan: @cpuguy83 i removed `ConsoleFromFile` completely, wdyt?

cpuguy83: I think if you do that it will never do TTY mode?

sozercan: tested and works with tty (output is glitchy for whatever reason on my terminal, but it was the same before)

cpuguy83 (approved):

PR #620, opened 2024-05-11
fix: added context to error

### LINKED ISSUE #396: [REQ] Provide more meaningful error messages

Adding more user context to the error messages that Copa returns helps troubleshoot unattended use of the tool. For example, I made a mistake to call copa with wrong reference to the image using digest like this:
```
myrepo/myimage:sha:<digest>
```
And the returned error was:
```
level=error msg="unexpected: TagNameOnly did create Tagged ref"
Error: unexpected: TagNameOnly did create Tagged ref
```
The above error doesn't help the user to understand what the problem is and leaves her wondering. 

Providing full context how the tool is invoked and what part of the invocation was wrong would be much friendlier.

prakrit55:
* added more context to the error message found while patching with wrong image reference

Closes #396

### DIFF
--- a/pkg/patch/patch.go
+++ b/pkg/patch/patch.go
@@ -85,7 +85,7 @@ func patchWithContext(ctx context.Context, ch chan error, image, reportFile, pat
 	}
 	taggedName, ok := imageName.(reference.Tagged)
 	if !ok {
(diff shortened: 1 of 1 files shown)


ashnamehrotra (approved):
LGTM, thanks!

PR #632, opened 2024-05-21
feat: add rpm package verification

### LINKED ISSUE #601: [QUESTION] Verify if there are format correctness check that need to be added given lack of support in rpmVer lib
### What is your question?

https://github.com/project-copacetic/copacetic/blob/d648155f5424a9f4cb13acd7209195846791873b/pkg/pkgmgr/rpm.go#L82

Turning copacetic TODO comments into issues from https://docs.google.com/spreadsheets/d/1XwNj1J6e2FrUhlqaIsV10l8_tgov7WodlkvpNZXYZMU/edit#gid=1386834576.

MiahaCybersec:
Closes #601

### DIFF
--- a/pkg/pkgmgr/rpm.go
+++ b/pkg/pkgmgr/rpm.go
@@ -83,10 +83,7 @@ func (st rpmDBType) String() string {
 // Depending on go-rpm-version lib for RPM version comparison rules.
 func isValidRPMVersion(v string) bool { // nolint:revive
 	err := isValidPackage(v)
-	if err != nil {
-		return false
-	}
-	return true
+	return err == nil
 }
 
 func isValidPackage(ver string) error {
--- a/pkg/pkgmgr/rpm.go
+++ b/pkg/pkgmgr/rpm.go
@@ -10,6 +10,7 @@ import (
 	"path/filepath"
 	"sort"
 	"strings"
+	"unicode"
 
 	"github.com/hashicorp/go-multierror"
 	rpmVer "github.com/knqyf263/go-rpm-version"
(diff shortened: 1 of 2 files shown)


pkg/pkgmgr/rpm.go
```diff
@@ -81,8 +82,22 @@ func (st rpmDBType) String() string {
 
 // Depending on go-rpm-version lib for RPM version comparison rules.
 func isValidRPMVersion(v string) bool { // nolint:revive
-	// TODO: Verify if there are format correctness check that need to be added given lack of support in rpmVer lib
-	return true
+	err := isValidPackage(v)
+	return err == nil
+}
+
+func isValidPackage(ver string) error {
```
ashnamehrotra, line 89: could we add a unit test to rpm_test file for this function?

pkg/pkgmgr/rpm.go
```diff
@@ -81,8 +82,22 @@ func (st rpmDBType) String() string {
 
 // Depending on go-rpm-version lib for RPM version comparison rules.
 func isValidRPMVersion(v string) bool { // nolint:revive
-	// TODO: Verify if there are format correctness check that need to be added given lack of support in rpmVer lib
-	return true
+	err := isValidPackage(v)
+	return err == nil
```
ashnamehrotra, line 86: instead of isValidPackage maybe we could leave it in this function since we are just checking version?

pkg/pkgmgr/rpm.go
```diff
@@ -81,8 +82,22 @@ func (st rpmDBType) String() string {
...
+func isValidPackage(ver string) error {
+	if !unicode.IsDigit(rune(ver[0])) {
+		return errors.New("upstream_version must start with digit")
+	}
+
+	allowedSymbols := ".-+~:_"
+	for _, s := range ver {
+		if !unicode.IsDigit(s) && !unicode.IsLetter(s) && !strings.ContainsRune(allowedSymbols, s) {
+			return errors.New("upstream_version includes invalid character")
+		}
```
ashnamehrotra, line 98: Using the strings.ContainsAny() function might be more concise here. Also maybe we can log the upstream_version in the error message. Otherwise the checks look good to me based on the docs!

ashnamehrotra (approved):
LGTM, thanks!

PR #635, opened 2024-05-25
ci: updated trivy dependency to v0.51.4

ChristofferNissen:
Replace trivy v0.45.1 with v.51.4 to resolve potential vulnerability with ECS.

Convert `report.Metadata.OS.Family` to string as it is a string under a custom struct type masking a string

Closes #397


go.mod
```diff
@@ -1,12 +1,12 @@
 module github.com/project-copacetic/copacetic
 
-go 1.21
+go 1.22.0
```
ashnamehrotra, line 4: could we specify this as just 1.22?

ChristofferNissen: Sure, this is what the cli put, but can change it

ashnamehrotra: it should be ok to run go mod tidy, is it resulting in an error?

ChristofferNissen: It just puts the value back to 1.22.0 in go.mod

ashnamehrotra (approved):
LGTM, thank you!

PR #641, opened 2024-05-29
ci: add CODECOV_TOKEN field to upload coverage

ashnamehrotra:

Closes #<_issue_ID_>


.github/workflows/build.yml
```diff
@@ -50,6 +50,8 @@ jobs:
         run: make test
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@125fc84a9a348dbcf27191600683ec096ec9021c # v4.4.1
+        env:
+          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
```
sozercan, line 54: We can't add this here, as forks do not have access to secrets. 

afaik we don't need codecov token for public repositories: https://about.codecov.io/blog/tokenless-uploads-for-github-actions/

sozercan (changes requested):

.github/workflows/build.yml
ashnamehrotra: it looks like they in the new release of codecov it is unsupported (https://github.com/codecov/codecov-action/releases/tag/v4.0.0), but forks will not need access?

MiahaCybersec: To quote one line of the CodeCov release that Ashna linked:

> Tokenless uploading is unsupported. However, PRs made from forks to the upstream public repos will support tokenless (e.g. contributors to OS projects do not need the upstream repo's Codecov token). This [doc](https://docs.codecov.com/docs/adding-the-codecov-token#github-actions) shows instructions on how to add the Codecov token.

So it looks like this repo can add the CodeCov token and it should be fine that nobody's forks will have access to it.

ashnamehrotra:
Added CODECOV_TOKEN to repo

sozercan (approved):
LGTM

PR #618, opened 2024-05-10
feat: Add error validation for update all

ashnamehrotra:
merge after #570 

Describe the changes in this pull request using active verbs such as _Add_, _Remove_, _Replace_ ...

Closes #593 

Part of #522

PR #644, opened 2024-05-30
feat: update all - check for upgradable packages

### LINKED ISSUE #594: Copa should also check if there are any package updates. If not, it should fast fail

ashnamehrotra:
Adds a check for upgradable packages before updating all when patching without scanner input.

Wanted feedback for microdnf check - discussed with @cpuguy83 to install dnf to call check-update option since microdnf does not support it, but the only way to remove it after is with `rpm -e dnf`.

Closes #594


cpuguy83:
For dnf based distros, have you tried using the tooling image and mounting the rootfs of the image you want to check?

Rough example:

```go
llb.Image(myToolingImage).Run(
  llb.AddMount(rootfsToCheck, "/tmp/rootfs"),
  llb.Args([]string{"sh", "-c", "dnf --installroot=/tmp/rootfs check-update"})
```

cpuguy83:
I say dnf, any distro that uses rpm+yum/dnf style repos and configs.

ashnamehrotra:
> For dnf based distros, have you tried using the tooling image and mounting the rootfs of the image you want to check?
> 
> Rough example:
> 
> ```go
> llb.Image(myToolingImage).Run(
>   llb.AddMount(rootfsToCheck, "/tmp/rootfs"),
>   llb.Args([]string{"sh", "-c", "dnf --installroot=/tmp/rootfs check-update"})
> ```

I made the changes to get this to work and was able to test it with yum/dnf based images. However, for microdnf based images (like quay.io/calico/cni:v3.15.1) it will always detect no upgrade when using dnf, even if the image is unpatched and would detect upgrades when running `microdnf install dnf; dnf check-update` in the container.

cpuguy83:
We investigated this a bit yesterday and it looks like `--installroot` is just broken with this case.
It loads the correct repos but does do the right thing at all and just always says that there are no updates.
I'm sure this could probably be fixed upstream in dnf/tdnf but that will probably take a long time.

The work-around we discussed is to solve an intermediate LLB state with dnf installed in the target image and run it like normal and write some file to the rootfs that we'll extract to determine if there are updates.

pkg/pkgmgr/rpm.go
```diff
@@ -434,6 +447,27 @@ func (rm *rpmManager) installUpdates(ctx context.Context, updates unversioned.Up
 	return &patchMerge, resultBytes, nil
 }
 
+func (rm *rpmManager) checkForUpgrades(ctx context.Context, toolPath string) error {
+	checkUpdateTemplate := `sh -c "set -x; %[1]s install dnf -y; dnf check-update -y; if [ $? -ne 0 ]; then echo >> /updates.txt; fi;"`
```
sozercan, line 451: can we use tdnf if it's already installed?

pkg/pkgmgr/rpm.go
```diff
@@ -401,22 +401,25 @@ func (rm *rpmManager) installUpdates(ctx context.Context, updates unversioned.Up
...
+		if !rm.checkForUpgrades(ctx, rm.rpmTools["yum"], checkUpdateTemplate) {
+			return nil, nil, fmt.Errorf("no patchable packages found")
 		}
 
 		const yumInstallTemplate = `sh -c '%[1]s upgrade %[2]s -y && %[1]s clean all'`
 		installCmd = fmt.Sprintf(yumInstallTemplate, rm.rpmTools["yum"], pkgs)
 	case rm.rpmTools["microdnf"] != "":
-		if rm.checkForUpgrades(ctx, rm.rpmTools["microdnf"]) != nil {
-			return nil, nil, errors.New("no upgradable packages")
+		checkUpdateTemplate := `sh -c "%[1]s install dnf -y; dnf check-update -y; if [ $? -ne 0 ]; then echo >> /updates.txt; fi;"`
```
sozercan, line 420: can we use the sh function added in #618

pkg/pkgmgr/rpm.go
```diff
@@ -401,22 +401,25 @@ func (rm *rpmManager) installUpdates(ctx context.Context, updates unversioned.Up
...
+		if !rm.checkForUpgrades(ctx, rm.rpmTools["dnf"], checkUpdateTemplate) {
+			return nil, nil, fmt.Errorf("no patchable packages found")
 		}
 
 		const dnfInstallTemplate = `sh -c '%[1]s upgrade %[2]s -y && %[1]s clean all'`
 		installCmd = fmt.Sprintf(dnfInstallTemplate, rm.rpmTools["dnf"], pkgs)
 	case rm.rpmTools["yum"] != "":
-		if rm.checkForUpgrades(ctx, rm.rpmTools["yum"]) != nil {
-			return nil, nil, errors.New("no upgradable packages")
+		checkUpdateTemplate := `sh -c 'if [ "$(%[1]s -q check-update | wc -l)" -ne 0 ]; then echo >> /updates.txt; fi'`
```
sozercan, line 412: same here

pkg/pkgmgr/rpm.go
```diff
@@ -401,22 +401,25 @@ func (rm *rpmManager) installUpdates(ctx context.Context, updates unversioned.Up
 	var installCmd string
 	switch {
 	case rm.rpmTools["dnf"] != "":
-		if rm.checkForUpgrades(ctx, rm.rpmTools["dnf"]) != nil {
-			return nil, nil, errors.New("no upgradable packages")
+		checkUpdateTemplate := `sh -c "%[1]s check-update; if [ $? -ne 0 ]; then echo >> /updates.txt; fi"`
```
sozercan, line 404: same here

pkg/pkgmgr/apk.go
```diff
@@ -165,6 +165,12 @@ func (am *apkManager) upgradePackages(ctx context.Context, updates unversioned.U
 	// TODO: Add support for custom APK config
 	apkUpdated := am.config.ImageState.Run(llb.Shlex("apk update"), llb.WithProxy(utils.GetProxy()), llb.IgnoreCache).Root()
 
+	// If updating all packages, check for upgrades before proceeding with patch
+	if updates == nil {
+		checkUpgradable := `sh -c "apk list 2>/dev/null | grep -q "upgradable" || exit 1"`
```
sozercan, line 170: same here

sozercan, line 170: can we use `apt list --upgradeable`?

sozercan, line 170: can we use `apk list --upgradeable`? this should be a smaller list

pkg/pkgmgr/rpm.go
ashnamehrotra: the command does not work unless we specify "sh -c" in the argument in addition to llb.Shlex. I am not sure why, but it is the same format followed in the installCmd.

pkg/pkgmgr/apk.go
ashnamehrotra: when testing with some images like `alpine:3.7.3`, `apk list `showed updates, and so did running `apk upgrade`. But running `apk list --upgradable` did not. The flag wasn't working as expected so it seemed apk list would be more accurate.

sozercan: Ok interesting, might be a bug with older version of apk

pkg/pkgmgr/rpm.go
sozercan: Ok let's follow up after if we can get this working

sozercan (approved):
LGTM. Thanks!

Issue #651, opened 2024-06-07
[REQ] support other tools in installToolsCmd for rpm tooling image


ashnamehrotra:
### What kind of request is this?

None

### What is your request or suggestion?

Would it be possible to support other tools (yum, microdnf etc) for installing busybox to our tooling image in rpm? Currently, we use [tdnf](https://github.com/project-copacetic/copacetic/blob/main/pkg/pkgmgr/rpm.go#L37), but if the user changes the tooling image via [source policies](https://project-copacetic.github.io/copacetic/website/faq#can-i-replace-the-package-repositories-in-the-image-with-my-own) there is a chance tdnf is not supported in that image.

### Are you willing to submit PRs to contribute to this feature request?

- [ ] Yes, I am willing to implement it.

Akash-Singh04:
Hey is this issue still open?If so,  I would like to work on it

Akash-Singh04:
Let me know if any separate changes are to be made for this issue or if it is resolved from the above PR as stated in #652

PR #670, opened 2024-06-19
feat: Added check for binaries to skip "installToolsCmd" 

### LINKED ISSUE #652: [REQ] Implement a possibility to skip "installToolsCmd"

New feature

I checked the official fedora and UBI 9 image, and there are some differences if and how `busybox cpio dnf-utils` are installable. So it would be great to implement the possibility to skip the installation of the tools and just check if the needed binaries are available. 
This would also be a fix for issue #651 

Akash-Singh04:
Updated probeRPMStatus to check if binaries are present and only if not present to all installToolsCmd
Closes #652 #651

### DIFF
--- a/pkg/pkgmgr/rpm.go
+++ b/pkg/pkgmgr/rpm.go
@@ -7,6 +7,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"os/exec"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


ashnamehrotra:
@Akash-Singh04 is there a specific image we can test the changes with?

PR #682, opened 2024-06-27
test: buildkit with defaults

### LINKED ISSUE #595: [REQ] Test buildkit with defaults

None

https://github.com/project-copacetic/copacetic/blob/d648155f5424a9f4cb13acd7209195846791873b/pkg/buildkit/buildkit_test.go#L218

Turning copacetic TODO comments into issues from https://docs.google.com/spreadsheets/d/1XwNj1J6e2FrUhlqaIsV10l8_tgov7WodlkvpNZXYZMU/edit#gid=1386834576.

MiahaCybersec:
Closes #595


MiahaCybersec:
It looks like the failing test is "default buildkit address", but the test runs and passes locally. Is it possible that the test is failing because GitHub actions doesn't have a locally running BuildKit client?

The test initializes `bkOpts` with no values, so `bkOpts.Addr` is an empty string when `NewClient` is called. It then tries to connect to a locally running BuildKit service (autoClient) which is where the errors are originating from.

@cpuguy83 mentioned at a previous LFX mentorship meeting that the old TODO is likely referring to testing BuildKit with nothing passed in, which is why this test does not perfectly match up with the old TODO in code.

ashnamehrotra:
@MiahaCybersec can you check what docker version is being used in the workflow? If it is not 23.0 or later, buildkit will not be enabled by default, and you would need to change the [unit test workflow ](https://github.com/project-copacetic/copacetic/blob/main/.github/workflows/build.yml#L30)to enable it

ashnamehrotra (approved):
LGTM

pkg/buildkit/buildkit_test.go
```diff
@@ -197,6 +197,18 @@ func TestNewClient(t *testing.T) {
 			cancel()
 			checkMissingCapsError(t, err, requiredCaps...)
 		})
+		t.Run("faulty addr", func(t *testing.T) {
+			t.Parallel()
+			addr := newMockBuildkitAPI(t) // this should return a faulty addr
```
cpuguy83, line 202: Can you have more details on this comment?
Its not clear from the call why this should be faulty and may produce something else in the future.

pkg/buildkit/buildkit_test.go
```diff
@@ -213,10 +225,16 @@ func TestNewClient(t *testing.T) {
 			err = ValidateClient(ctxT, client)
 			assert.NoError(t, err)
 		})
+		t.Run("default buildkit addr", func(t *testing.T) {
+			t.Parallel()
+			bkOpts := Opts{} // Initialize with default values
+			client, err := NewClient(context.Background(), bkOpts)
```
cpuguy83, line 231: Can you use `context.TODO()` for these?

cpuguy83 (changes requested):

pkg/buildkit/buildkit_test.go
MiahaCybersec: That comment was left in by mistake, the address validation in `client.New` does not properly parse invalid addresses. If an address such as `300.300.300.300` is passed in, it shows no errors.

I've pushed a new commit that updates the test to be named `faulty key path` and added additional error validation for the tests.

MiahaCybersec:
I'm not sure why the lint is failing. It says `platforms.Normalize` is deprecated, but then says to use `platforms.Normalize`.

sozercan:
@MiahaCybersec haha that's confusing. Looks like it's deprecated in favor of `platforms.Normalize` in `github.com/containerd/platforms`. If you change the import to `github.com/containerd/platforms` it should work.

MiahaCybersec:
The unit tests for `GetDockerTransport` have been removed because they were incomplete. The tests did not include all of the necessary checks, and mocks would be required to properly test them. The unit tests will be added in a separate pull request to increase test coverage.

cpuguy83 (approved):

PR #686, opened 2024-07-03
ci: remove centos eol img

ashnamehrotra:

Closes #<_issue_ID_>


sozercan (approved):
LGTM. Looks like this went EOL few days ago.

`14: curl#6 - "Could not resolve host: mirrorlist.centos.org; Unknown error"`
https://www.redhat.com/en/topics/linux/centos-linux-eol

Issue #600, opened 2024-05-02
[REQ] Add support for custom APT config and gpg key injection


ashnamehrotra:
### What kind of request is this?

None

### What is your request or suggestion?

https://github.com/project-copacetic/copacetic/blob/d648155f5424a9f4cb13acd7209195846791873b/pkg/pkgmgr/dpkg.go#L239C1-L240C1

Turning copacetic TODO comments into issues from https://docs.google.com/spreadsheets/d/1XwNj1J6e2FrUhlqaIsV10l8_tgov7WodlkvpNZXYZMU/edit#gid=1386834576.

### Are you willing to submit PRs to contribute to this feature request?

- [ ] Yes, I am willing to implement it.

h4l0gen:
Hii @ashnamehrotra, In first look we want to extend user's control by giving them functionality for custom APT config file. 
If you have something in your mind regarding this which can help me to solve this issue, please tell me. So we can work on this in proper way.

ashnamehrotra:
Hi @h4l0gen, based on the comment, I think we should set up the configuration within the original `dm.config.ImageState` prior to applying apt update on the resulting state based custom config input into Copa.

Akash-Singh04:
Hey is this issue still open?If so,  I would like to work on it

ashnamehrotra:
Hi @h4l0gen, are you still planning on working on this issue?

h4l0gen:
Hi @ashnamehrotra @Akash-Singh04 I am really out of time right now. Akash please feel free to work on it. I will look into problem of my solution if i found time 🙂.

Akash-Singh04:
@ashnamehrotra kindly review and let me know if any more changes are required

PR #690, opened 2024-07-04
feat: Added support for custom APT config and gpg key injection

SkySingh04:
Add functionality to make the `customAPTConfig` and `gpgKeys` parameters optional in the `installUpdates` function of the `dpkgManager` struct. The function now checks if these parameters are provided before applying the custom APT configuration and adding the GPG keys.

-  Conditional checks for `customAPTConfig` and `gpgKeys` in the `installUpdates` function.
-  Update the function to apply configurations only if the parameters are provided.

**Closes:** #600


ashnamehrotra:
@Akash-Singh04 I'm not sure I follow. How are these parameters being set from the client before they are passed down via installUpdates?

PR #706, opened 2024-07-17
feat: add oracle support

### LINKED ISSUE #190: [REQ] Add Oracle Linux as supported OS

New feature

Currently, Oracle Linux (https://hub.docker.com/_/oraclelinux) cannot be patched and the following error message is shown:
```text
Error: unsupported osType oracle specified
```
Please, add support for Oracle Linux.

Note: Oracle Linux 7 uses `yum`, while starting with version 8 `dnf` is used as package manager.

MiahaCybersec:
Closes #190 

Did we want to add some docs that Oracle is only supported without a vulnerability scan?


sozercan:
can we fast fail if scanner input is specific with a descriptive error message?

yes please add to docs. thanks!

website/docs/troubleshooting.md
```diff
@@ -2,6 +2,24 @@
...
+patch -r /oracle-7.9-vulns.json -i docker.io/library/oraclelinux:7.9 --ignore-errors
+```
+
+Without a vulnerability scan, Copa will update all packages in the image:
+
+```bash
+patch -i docker.io/library/oraclelinux:7.9
+```
+
+Oracle reports CVEs in a way that causes Trivy and Copa to report false positives. To patch the entire image, use the Copa `--ignore-errors` flag or don't pass in a vulnerability scan. See [this GitHub issue](https://github.com/aquasecurity/trivy/issues/1967#issuecomment-1092987400) for more information.
```
ashnamehrotra, line 21:
```suggestion
Oracle reports CVEs in a way that causes Trivy to report false positives that Copa will be unable to patch. To patch the entire image, use the Copa `--ignore-errors` flag or omit the vulnerability scan report to upgrade all outdated packages. See [this GitHub issue](https://github.com/aquasecurity/trivy/issues/1967#issuecomment-1092987400) for more information.
```

integration/patch_test.go
```diff
@@ -110,9 +114,13 @@ func TestPatch(t *testing.T) {
 			}
 
 			// currently validation is only present when patching with a scan report
-			if reportFile {
+			if reportFile && !strings.Contains(img.Image, "oracle") {
```
ashnamehrotra, line 117: in addition to patch_test, can we fast fail when running copa patch command too?

integration/patch_test.go
```diff
@@ -189,6 +197,12 @@ func patch(t *testing.T, ref, patchedTag, path string, ignoreErrors bool, report
 		reportPath = "-r=" + path + "/scan.json"
 	}
 
+	if strings.Contains(ref, "oracle") && reportFile {
```
MiahaCybersec, line 200: There is a fast fail in the tests if Oracle is being patched with a Trivy scan report. Did you want it to fast fail when an end user passes in an Oracle image during normal Copa usage as well?

ashnamehrotra: yes, since we will not be able to patch it we can fast fail with a error message to either ignore errors or update all

MiahaCybersec: Rather than fast failing, I've added some checks so that if an Oracle image is passed in and ignore errors is false, `log.Info()` says an Oracle image was detected and ignore errors is being set to true. Doing it this way will mean that even if a user doesn't read the docs on Oracle support, Copa still handles the image type properly without needing to exit.

ashnamehrotra: @MiahaCybersec I think it would be better to let the user decide and re-patch if they want --ignore-errors set to true? Since we would want them to be aware of the issue/read the docs because if they re-scan the image they would see the vulns

pkg/pkgmgr/rpm.go
```diff
@@ -191,7 +191,14 @@ func (rm *rpmManager) InstallUpdates(ctx context.Context, manifest *unversioned.
 	var updates unversioned.UpdatePackages
 	var rpmComparer VersionComparer
 	var err error
+
 	if manifest != nil {
+		if manifest.Metadata.OS.Type == "oracle" && !ignoreErrors {
+			err := fmt.Errorf("Detected Oracle image being passed in without ignore errors \n" +
+				"Please read https://project-copacetic.github.io/copacetic/website/troubleshooting before patching your Oracle image")
```
ashnamehrotra, line 198:
```suggestion
		if manifest.Metadata.OS.Type == "oracle" && !ignoreErrors {
			err := fmt.Errorf("Detected Oracle image passed in \n" +
				"Please read https://project-copacetic.github.io/copacetic/website/troubleshooting before patching your Oracle image")
```

ashnamehrotra:
@MiahaCybersec for codecov coverage, maybe we can modify the getOSType() to check for Oracle?

integration/patch_test.go
```diff
@@ -189,6 +197,12 @@ func patch(t *testing.T, ref, patchedTag, path string, ignoreErrors bool, report
...
+		err := errors.New("oracle is not supported with trivy scans")
+		fmt.Println(err)
```
ashnamehrotra, line 202: instead of this, lets test "Error: Detected Oracle image passed in" since copa fast fails anyways, this will help with codecov

integration/patch_test.go
```diff
@@ -248,7 +259,12 @@ func (s *scannerCmd) scan(t *testing.T, ref string, ignoreErrors bool) {
 	cmd.Env = append(cmd.Env, os.Environ()...)
 	cmd.Env = append(cmd.Env, dockerDINDAddress.env()...)
 	out, err := cmd.CombinedOutput()
-	assert.NoError(t, err, string(out))
+
+	if strings.Contains(ref, "oracle") {
+		assert.Contains(t, string(out), "Detected OS: oracle")
```
ashnamehrotra, line 264: why are we changing the scanner test?

website/docs/troubleshooting.md
```diff
@@ -2,6 +2,23 @@
...
 ---
 
+## Copa and Trivy throw errors when Oracle Linux is passed in
+
+Copa supports patching Oracle Linux in two ways:
+
+With a vulnerability scan, `--ignore-errors` must be passed in:
+
+```bash
+patch -r /oracle-7.9-vulns.json -i docker.io/library/oraclelinux:7.9 --ignore-errors
```
sozercan, line 12: copa patch?

website/docs/troubleshooting.md
```diff
@@ -2,6 +2,23 @@
...
+```
+
+Without a vulnerability scan, Copa will update all packages in the image:
+
+```bash
+patch -i docker.io/library/oraclelinux:7.9
```
sozercan, line 18: copa patch?

website/docs/troubleshooting.md
sozercan: i don't think this is going to patch all vulns, is that right? can we add a note?

MiahaCybersec: This will patch all vulnerabilities, but `--ignore-errors` must be passed in due to how Oracle handles CVEs.

Even if an Oracle image is fully patched, Trivy scans will complain about `_fips`. This is known upstream but has unfortunately remained unfixed for a couple years.

website/docs/troubleshooting.md
MiahaCybersec: This would be referencing `copa patch`

ashnamehrotra: lets change it to copa patch to be consistent with other parts of the doc and avoid confusion

website/docs/troubleshooting.md
ashnamehrotra: we can add a note that this will patch all errors aside from the false positives due to Oracle CVE

ashnamehrotra (approved):
LGTM

PR #732, opened 2024-08-05
refactor: patch with context

### LINKED ISSUE #723: [REQ] Refactor patchWithContext()

Refactor patchWithContext() and make it more modular.

_No response_

MiahaCybersec:
Closes #723 

This PR primarily addresses the untestability of the core patch function within Copa. By breaking the code into smaller blocks, the core functionality of Copa can more easily be tested and extended in the future, and bugs should be easier to find.

### DIFF
--- a/pkg/patch/patch.go
+++ b/pkg/patch/patch.go
@@ -221,7 +221,8 @@ func patchWithContext(ctx context.Context, ch chan error, image, reportFile, use
 	})
 
 	eg.Go(func() error {
-		if err := dockerLoad(ctx, pipeR); err != nil {
+		err = dockerLoad(ctx, pipeR)
+		if err != nil {
 			return err
 		}
 		return pipeR.Close()
@@ -358,7 +359,7 @@ func generatePatchedTag(dockerNormalizedImageName reference.Named, userSuppliedP
 	}
 
 	if userSuppliedPatchTag != "" {
-		copaTag = fmt.Sprintf("%s-%s", officialTag, userSuppliedPatchTag)
+		copaTag = userSuppliedPatchTag
 		return copaTag
 	} else if officialTag == "" {
 		log.Warnf("No output tag specified for digest-referenced image, defaulting to `%s`", defaultPatchedTagSuffix)
--- a/pkg/patch/patch.go
+++ b/pkg/patch/patch.go
@@ -42,7 +42,7 @@ const (
 	defaultTag              = "latest"
 )
 
-type BuildOpts struct {
+type TrivyOpts struct {
 	BkClient                  *client.Client
 	SolveOpt                  *client.SolveOpt
 	Image                     string
@@ -55,6 +55,11 @@ type BuildOpts struct {
 	DockerNormalizedImageName reference.Named
 	PatchedImageName          string
 	Format                    string
+}
+
+type BkClient struct {
+	BkClient *client.Client
+	SolveOpt *client.SolveOpt
 }
 
 type BuildStatus struct {
@@ -196,7 +201,7 @@ func patchWithContext(ctx context.Context, ch chan error, image, reportFile, use
 	eg.Go(func() error {
 		err = buildkitBuild(
 			BuildContext{ctx},
-			&BuildOpts{
+			&TrivyOpts{
 				bkClient, &solveOpt, image, ch,
 				reportFile, workingFolder, updates, ignoreError,
 				output, dockerNormalizedImageName, patchedImageName, format,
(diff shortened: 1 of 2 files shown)


pkg/patch/patch.go
```diff
@@ -166,113 +180,14 @@ func patchWithContext(ctx context.Context, ch chan error, image, reportFile, pat
...
-					}
-				}
-			}
-
-			return res, nil
-		}, buildChannel)
-
+		err = buildkitBuild(
+			BuildContext{ctx},
+			&TrivyOpts{
```
RobertKielty, line 185: I like that you've collated the TrivyOpts into a struct, but I'm not sure why there's a referecnce to a bkClient in there? 

When I think TrivyOpts I automatically think of the command line options that are passed into trivy, when I review those options there's no (obvious) reference to BuildKit; so if TrivyOpt is _not_ modelling commandline options, then maybe it should be re-named to better reflect its role here?

pkg/patch/patch_test.go
```diff
@@ -291,3 +298,228 @@ func TestGetOSVersion(t *testing.T) {
 		})
 	}
 }
+
+func TestGeneratePatchedTag(t *testing.T) {
```
RobertKielty, line 302: I like this test! 

It is easy to understand from having worked with you this.

That test names could be even more explicit so as to differentiate between two types of tag that we are considering here in this test.

We have two tags

1. The user-supplied tag which leaves a breadcrumb indicating that the image was patched by copa 
2. The imagename tag which appears at the end of the image

If you use `ImageTag` and `UserSuppliedTag` in the test names a new contributor (or a curious end-user) would more easily understand this test.

pkg/patch/patch_test.go
```diff
@@ -291,3 +298,228 @@ func TestGetOSVersion(t *testing.T) {
...
+			named, _ := reference.ParseNormalizedNamed(tc.dockerImageName)
+			patchedTag := generatePatchedTag(named, tc.userSuppliedPatchTag)
+			if patchedTag != tc.expectedPatchedTag {
+				t.Errorf("expected patchedTag to be %s but got %s", tc.expectedPatchedTag, patchedTag)
+			}
+		})
+	}
+}
+
+func TestUpdateManifest(t *testing.T) {
```
RobertKielty, line 340: In order to understand this test I need a definition of what an "errpackage" is. Could be done via a comment or a changed variable name.

Also, I'm struggling to understand what unversioned means here, (did some _very quick_ digging around the types associated with the code under test but as I was missing your guidance I could figure it out myself! Also, I am about to head out the door so didn't have the time to pour over the code)

Use this test to explain what is going on here. Don't change the underlying code, just provide some short commentary here on what is happening,

The audience for the commentary is between an end-user and a new contributor, assume the audience knows how to use copa and that this test gives them crucial insight in to how copa behaves.

RobertKielty: Other than that, the test is good, I can kind of see what is going on. It's just the signage that needs improving here.

RobertKielty (changes requested):
@MiahaCybersec I have submitted some review comments on the tests and code. 

This is good work that will make it easier for the Copa project to on-board new contributors, well done!

My requested changes are largely around improving the story telling around how copa works which will benefit both early adopting end-users and new contributors.

I have one question on the TrivyOpts and why that include non-trivy related fields (there's a bk client in there) 

The most important thing to do with work like this is to test it extensively and ensure that no behaviour has changed as a result of the submitted changes.

pkg/patch/patch.go
```diff
@@ -74,35 +97,26 @@ func removeIfNotDebug(workingFolder string) {
...
-		}
+		dockerNormalizedImageName = reference.TagNameOnly(dockerNormalizedImageName)
 	}
-	_, err = reference.WithTag(imageName, patchedTag)
+
+	patchedTag := generatePatchedTag(dockerNormalizedImageName, userSuppliedPatchTag)
+
+	_, err = reference.WithTag(dockerNormalizedImageName, patchedTag)
 	if err != nil {
 		return fmt.Errorf("%w with patched tag %s", err, patchedTag)
```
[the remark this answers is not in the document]
MiahaCybersec: The output of this would look like the following:

`Error: invalid tag format with patched tag :something!`

I don't think this error needs much changed?

pkg/patch/patch_test.go
MiahaCybersec: In the test case we are only supplying a user patched tag, which can either be empty `""`, or supplied `"example-supplied-tag"`. The differentiation is done within the function itself, not the parameters being passed in.

This is why for some of the test cases, an empty string is passed in but the expected result contains `defaultPatchedTagSuffix`

pkg/patch/patch.go
```diff
@@ -74,35 +97,26 @@ func removeIfNotDebug(workingFolder string) {
...
+// patchWithContext patches the user-supplied image, image.
+func patchWithContext(ctx context.Context, ch chan error, image, reportFile, userSuppliedPatchTag, workingFolder, scanner, format, output string, ignoreError bool, bkOpts buildkit.Opts) error {
+	dockerNormalizedImageName, err := reference.ParseNormalizedNamed(image)
 	if err != nil {
 		return err
 	}
-	if reference.IsNameOnly(imageName) {
+
+	if reference.IsNameOnly(dockerNormalizedImageName) {
 		log.Warnf("Image name has no tag or digest, using latest as tag")
```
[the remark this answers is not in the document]
MiahaCybersec: This error has been updated to display the following warn which should be a lot clearer to end users:

`WARN[0000] Image name docker.io/library/alpine has no tag or digest, defaulting to docker.io/library/alpine:latest `

pkg/patch/patch.go
```diff
@@ -42,6 +42,32 @@ const (
 	defaultTag              = "latest"
 )
 
+type TrivyOpts struct {
+	Image                     string
+	Ch                        chan error
```
ashnamehrotra, line 47: I'm not sure if it makes sense to have channel in TrivyOpts. Also, so we aren't making this specific to Trivy, we could call it scannerOpts? We would still want this to be compatible with https://github.com/project-copacetic/scanner-plugin-template

pkg/patch/patch_test.go
```diff
@@ -291,3 +299,236 @@ func TestGetOSVersion(t *testing.T) {
...
+		},
+		{
+			name:                 "WithTag_NoUserSupplied",
+			dockerImageName:      "docker.io/redhat/ubi9:latest",
+			userSuppliedPatchTag: "",
+			expectedPatchedTag:   fmt.Sprintf("latest-%s", defaultPatchedTagSuffix),
+		},
+		{
+			name:                 "WithTag_UserSupplied",
+			dockerImageName:      "docker.io/librari/ubuntu:jammy-20231004",
```
ashnamehrotra, line 332: could we also add a test for when we reference the image with a digest?

pkg/patch/patch_test.go
```diff
@@ -291,3 +299,236 @@ func TestGetOSVersion(t *testing.T) {
...
+		name      string
+		pipeR     io.Reader
+		mockCmd   *exec.Cmd
+		expectErr bool
+	}{
+		{
+			name:      "Unrecognized image format",
+			pipeR:     &mockReader{nil, errors.New("unrecognized image format")},
+			mockCmd:   exec.Command("echo", "test"),
+			expectErr: true,
```
ashnamehrotra, line 512: would it be possible to add a test case expectErr is false?

pkg/patch/patch.go
```diff
@@ -301,6 +231,147 @@ func patchWithContext(ctx context.Context, ch chan error, image, reportFile, pat
...
+		if !slices.Contains(errPkgs, update.Name) {
+			validatedManifest.Updates = append(validatedManifest.Updates, update)
+		}
+	}
+	return validatedManifest
+}
+
+func generatePatchedTag(dockerNormalizedImageName reference.Named, userSuppliedPatchTag string) string {
+	// officialTag is typically the versioning tag of the image as published in a container registry
+	var officialTag string
```
ashnamehrotra, line 351: maybe we can rename this to be currentTag?

ashnamehrotra (changes requested):

pkg/patch/patch_test.go
MiahaCybersec: When `dockerLoad()` is called, it takes in a TAR header as a parameter and I was unable to find a way to extract this data to create a successful test. Looking at it in the debugger, there's no clear data for me to extract and put into the tests either.

If you or Sertaç happen to know a valid TAR header or how we can access one, I'll happily add a new test case 😄

ashnamehrotra:
LGTM, will wait to merge. @MiahaCybersec @RobertKielty and I discussed adding canary testing before merging large changes.

MiahaCybersec:
Changing PR to draft until it gets further review 👍

MiahaCybersec:
Any updates on internal testing for this refactoring PR?

MiahaCybersec:
This PR passes all tests, except for the 4 that are currently failing due to issue #877. Otherwise, it is functionally ready for review by maintainters.

PR #619, opened 2024-05-11
feat: Add guardrails for parsing OS version

h4l0gen:
adding guardrails for parsing OS version as per discussion 

Closes #438

PR #736, opened 2024-08-12
docs: add tooling image versioning docs

### LINKED ISSUE #437: [DOC] clarify OS Version in scanner
### What kind of documentation improvement is needed?

Missing information or guidance

### What is the change that is needed?

Clarify this info is used to find the appropriate tooling image for the OS. This applies to deb and rpm based images today.
For example, for debian, it'll discard the minor version and use the slim image with the corresponding major version (example: debian:12-slim)
for Ubuntu, it'll use the major.minor. Since there's no major.minor.patch image for ubuntu, that's why it's failing in your example below.
for RPM, for Azure linux, we parse X.Y.Z (for example, "2.0.20231130") but discard Z (which ends up with mcr.microsoft.com/cbl-mariner/base/core:2.0). for non-Azure Linux based RPM, just use mcr.microsoft.com/cbl-mariner/base/core:2.0 today.

### Are you willing to submit PRs to contribute to documentation?

- [X] Yes, I am willing to implement it.

### LINKED ISSUE #438: [REQ] add guardrails for parsing OS version

For example, for ubuntu if a scanner return X.Y.Z, copa tries to pull in the ubuntu:X.Y.Z image which doesn't exist

related to #437 

- [X] Yes, I am willing to implement it.

MiahaCybersec:
Closes #437, #438, #619


pkg/patch/patch.go
```diff
@@ -341,7 +341,15 @@ func getOSVersion(ctx context.Context, osreleaseBytes []byte) (string, error) {
...
 
-	return osData["VERSION_ID"], nil
+	var osVersion string
+	semanticVersion := strings.Split(osData["VERSION_ID"], ".")
+	switch {
+	case len(semanticVersion) > 2:
+		osVersion = strings.Join(semanticVersion[:2], ".")
+	default:
+		osVersion = osData["VERSION_ID"]
+	}
```
ashnamehrotra, line 351: only changing the osVersion here won't work since the approach is different for each pkgmanager as listed in #437

MiahaCybersec:
The revert commits were rebased to be signed off, but the DCO check is still failing due to the previous unsigned commits still being in the commit history

website/docs/faq.md
```diff
@@ -17,6 +17,20 @@ To patch vulnerabilities for applications, you can package these applications an
...
+
+Debian: All debian images have their `minor.patch` versioning stripped and `-slim` appended. e.g. if `debain:12.6` is passed in, `debian:12-slim` is used as the tooling image.
+
+Ubuntu: All Ubuntu images use the same versioning that was passed in. e.g. if `ubuntu:24.04` is passed in, `ubuntu:24.04` will be used for the tooling image.
+
+There is one caveat for Ubuntu images. If an Ubuntu image is being patched without a Trivy scan, Copa is unable to parse a scan for versioning information. In these scenarios, Copa will fallback to `debian:stable-slim` as the tooling image.
+
+RPM: All RPM-based images will use `mcr.microsoft.com/cbl-mariner/base/core:2.0` as the tooling image.
+
+APK: APK-based images never use a tooling image.
```
ashnamehrotra, line 32:
```suggestion
APK: APK-based images never use a tooling image, as Copa does not patch distroless alpine images.
```

website/docs/faq.md
```diff
@@ -17,6 +17,20 @@ To patch vulnerabilities for applications, you can package these applications an
 
 If you find that your storage is rapidly being taken up after working with Copa, run `docker system prune`. This will prune all unused images, containers and caches. 
 
+## How does Copa determine what tooling image to use?
+
+All images being passed into Copa has their versioning data carefully stripped so that an appropriate tooling image can be obtained from a container repository.
+
+Debian: All debian images have their `minor.patch` versioning stripped and `-slim` appended. e.g. if `debain:12.6` is passed in, `debian:12-slim` is used as the tooling image.
+
+Ubuntu: All Ubuntu images use the same versioning that was passed in. e.g. if `ubuntu:24.04` is passed in, `ubuntu:24.04` will be used for the tooling image.
```
ashnamehrotra, line 26: were we able to confirm this with a test?

website/docs/faq.md
```diff
@@ -17,6 +17,20 @@ To patch vulnerabilities for applications, you can package these applications an
 
 If you find that your storage is rapidly being taken up after working with Copa, run `docker system prune`. This will prune all unused images, containers and caches. 
 
+## How does Copa determine what tooling image to use?
+
+All images being passed into Copa has their versioning data carefully stripped so that an appropriate tooling image can be obtained from a container repository.
```
ashnamehrotra, line 22:
```suggestion
All images being passed into Copa have their versioning data carefully stripped so that an appropriate tooling image can be obtained from a container repository.
```

ashnamehrotra:
@MiahaCybersec instead of using the OS images themselves in the examples, maybe we can use different image names to avoid confusion?

ashnamehrotra:
> The revert commits were rebased to be signed off, but the DCO check is still failing due to the previous unsigned commits still being in the commit history

It looks like commit message `[d4fd6d8](https://github.com/project-copacetic/copacetic/pull/736/commits/d4fd6d8af5cc7cb19658e5e84e5c577a93601891)` needs to be modified to include sign off

MiahaCybersec:
Had to force push a bunch of commits to correct the previously unsigned commits

website/docs/faq.md
MiahaCybersec: I was able to on my machine. Here's the relevant code for that determination. If no scan is passed in, manifest is nil and we'll hit the else block for Ubuntu images to use Ubuntu tooling images. If the manifest is nil or it contains Debian, we default to Debian for tooling images.

image

ashnamehrotra: it looks like we are setting osType to Debian always?

ashnamehrotra (approved):
LGTM, thank you!

PR #741, opened 2024-08-15
feat: skip install tools cmd

### LINKED ISSUE #651: [REQ] support other tools in installToolsCmd for rpm tooling image

MiahaCybersec:
Closes #651, #652


MiahaCybersec:
Currently working on fixing the failing test

pkg/pkgmgr/rpm.go
```diff
@@ -361,6 +367,52 @@ func (rm *rpmManager) probeRPMStatus(ctx context.Context, toolImage string) erro
...
+	}
+
+	// missingTools indicates which tools, if any, need to be installed within the tooling image
+	var missingTools []string
+	requiredToolingList := []string{"busybox", "dnf-utils", "cpio"}
+
+	for i := range requiredToolingList {
+		found := false
+		for x := range applicationsListSplit {
+			if applicationsListSplit[x] == requiredToolingList[i] {
```
ashnamehrotra, line 399: maybe we can call a contains function here to simplify code?

ashnamehrotra: contains did not work, marking as resolved

ashnamehrotra:
@MiahaCybersec when testing this locally, it does work with a cbl-mariner to Fedora conversion source policy and doesn't work with Redhat conversion (as expected when patching a mariner image). However, I still see "dnf install busybox dnf-utils cpio -y" in the logs with Fedora, which should already have busybox?

pkg/patch/patch.go
```diff
@@ -163,6 +163,20 @@ func patchWithContext(ctx context.Context, ch chan error, image, reportFile, pat
 		return err
 	}
 
+	if solveOpt.SourcePolicy != nil {
+		switch {
+		case strings.Contains(solveOpt.SourcePolicy.Rules[0].Updates.Identifier, "redhat"):
+			err = errors.New("RedHat is not supported via source policies due to busybox not being in the RHEL repos\n" +
```
ashnamehrotra, line 169:
```suggestion
			err = errors.New("RedHat is not supported via source policies due to BusyBox not being in the RHEL repos\n" +
```

pkg/patch/patch.go
```diff
@@ -163,6 +163,20 @@ func patchWithContext(ctx context.Context, ch chan error, image, reportFile, pat
...
+				"Please use a different RPM-based image")
+			return err
+
+		case strings.Contains(solveOpt.SourcePolicy.Rules[0].Updates.Identifier, "rockylinux"):
+			err = errors.New("RockyLinux is not supported via source policies due to busybox not being in the RockyLinux repos\n" +
```
ashnamehrotra, line 174:
```suggestion
			err = errors.New("RockyLinux is not supported via source policies due to BusyBox not being in the RockyLinux repos\n" +
```

MiahaCybersec:
BusyBox is not preinstalled on Fedora, but it is available in the Fedora repos. However, BusyBox is not available in the RHEL or Rocky repos.

pkg/pkgmgr/rpm.go
```diff
@@ -361,6 +367,52 @@ func (rm *rpmManager) probeRPMStatus(ctx context.Context, toolImage string) erro
...
+	// packageManagersInstalled is the package manager(s) available within the tooling image
+	// RPM must be excluded from this list as it cannot connect to RPM repos
+	var packageManagersInstalled []string
+	packageManagerList := []string{"tdnf", "dnf", "microdnf", "yum"}
+
+	for i := range applicationsListSplit {
+		for _, packageManager := range packageManagerList {
+			if applicationsListSplit[i] == packageManager {
+				packageManagersInstalled = append(packageManagersInstalled, applicationsListSplit[i])
+			}
```
ashnamehrotra, line 388: this could potentially be a very large for loop if we are iterating each application and then each package manager for each application. Instead of iterating each application, is there a way we can simplify the applicationsListSplit string with regexp to see what matches tdnf/dnf/microdnf/yum ?

ashnamehrotra: if that doesn't work we could also create a map by iterating applicationsListSplit once and map those applications as "true" value and check for those when iterating packageManagerList

MiahaCybersec:
Only failing check is codecov/project - presumably a similar case to #706

ashnamehrotra (approved):

PR #773, opened 2024-09-15
chore: update go.mod changed go version from 1.22 to 1.23

ishwarmt:

Closes #<_issue_ID_>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,8 +1,8 @@
 module github.com/project-copacetic/copacetic
 
-go 1.22.0
+go 1.23.0
 
-toolchain go1.22.1
+toolchain go1.23.1
 
 require (
 	github.com/aquasecurity/trivy v0.55.0


ashnamehrotra:
@ishwarmt running a go mod tidy may fix the dependency failure

go.mod
```diff
@@ -1,8 +1,8 @@
 module github.com/project-copacetic/copacetic
 
-go 1.22.0
+go 1.23.0
```
sozercan, line 3: afaik this signals that 1.23 is the min version require to import this. copa isn't designed to be use as a library and not sure why would we want 1.23 as a min version?

sozercan (changes requested):

sozercan:
closing this, feel free to re-open if there's a specific reason we want to bump go directive

PR #928, opened 2025-02-26
Fix openvex report oci id bug

### LINKED ISSUE #667: [BUG] fix openvex report oci id
### Version of copa

_No response_

### Expected Behavior

it should be:

```
"@id": "pkg:oci/nginx@sha256....",
```

https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#oci

However, I am not sure if we can do this since image is not at the registry at this time so sha256 will not be accurate. tag might be more suitable?

### Actual Behavior

vex statement output has

```
"@id": "pkg:oci/docker.io/library/nginx:1.21.6-patched",
```

https://project-copacetic.github.io/copacetic/website/output

### Steps To Reproduce

_No response_

robert-cronin:
# Summary of changes
Adds logic to qualify the actual SHA of the patched image
Changes patch.go to uses the correct pURL format (e.g. nginx@sha256:...) for the namespace for oci container names
Falls back to patchedImageName if digest not found.

Fixes #667 

# Validation
Vex report output from a test run on nginx, the tail of the generated vex report:
image


sozercan:
i think linter failure is due to incompat between go 1.24 and golang-ci lint version. opened #930 to bump linter

pkg/patch/patch.go
```diff
@@ -386,3 +404,29 @@ func dockerLoad(ctx context.Context, pipeR io.Reader) error {
 
 	return cmd.Run()
 }
+
+// e.g. "docker.io/library/nginx:1.21.6-patched".
+func getRepoNameWithDigest(ctx context.Context, patchedImageName string) (string, error) {
+	cmd := execCommandContext(ctx, "docker", "image", "inspect",
```
sozercan, line 410: can we do this without docker cli?

robert-cronin: I’m wondering if we could make use of the BuildKit exporter response, since it should include the digest of the patched image. The challenge is that this information only becomes available once we’ve finished generating the VEX report, because everything happens within the same build command. I’d really appreciate any tips or guidance on how to make that work!

cpuguy83: Do we need to generate the vex from within the build?

ashnamehrotra (approved):
LGTM

PR #946, opened 2025-03-06
feat: mirror tooling images to ghcr

### LINKED ISSUE #937: [REQ] mirror tooling images to ghcr

None

dockerhub introduced new throttling restrictions, and we have seen a few user reports that they are getting throttled due to tooling images coming from docker hub

while users can add buildkit source policies to convert these tooling images into non-dockerhub locations, it is not a great user experience that they have to do this.

we should mirror tooling images into ghcr to avoid this issue. since we might hit docker hub limits ourselves, mirror.gcr.io pull through cache might be a good alternative

robert-cronin:
Adds workflow that will mirror tooling images to ghcr
Changes the tooling image references to ghcr

Closes #937

## Verification
for the cache pulling: https://github.com/robert-cronin/copacetic/actions/runs/13689645432/job/38280415279

### DIFF
--- a/.github/workflows/mirror-tooling-images.yml
+++ b/.github/workflows/mirror-tooling-images.yml
@@ -1,0 +1,58 @@
+name: "[Daily] Mirror Tooling Images to GHCR"
+
+on:
+  schedule:
+    - cron: "0 3 * * *" # every day at 3 AM UTC
+  workflow_dispatch:
+    inputs:
+      ghcr_namespace:
+        description: "GHCR Namespace"
+        required: true
+        default: "ghcr.io/project-copacetic/copacetic"
+
+env:
+  GHCR_NAMESPACE: ${{ github.event.inputs.ghcr_namespace || 'ghcr.io/project-copacetic/copacetic' }}
+
+jobs:
+  mirror:
+    runs-on: ubuntu-latest
+    permissions:
+      packages: write
+      contents: read
+
+    strategy:
+      matrix:
+        images:
+          - "mirror.gcr.io/debian:10-slim"
+          - "mirror.gcr.io/debian:11-slim"
+          - "mirror.gcr.io/debian:12-slim"
+          - "mirror.gcr.io/debian:stable-slim"
+          - "mirror.gcr.io/ubuntu:20.04"
+          - "mirror.gcr.io/ubuntu:22.04"
+          - "mcr.microsoft.com/azurelinux/base/core:3.0"
+          - "mcr.microsoft.com/cbl-mariner/base/core:2.0"
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@e3d2460bbb42d7710191569f88069044cfb9d8cf # v4.2.2
+
+      - name: Install oras CLI
+        uses: oras-project/setup-oras@5c0b487ce3fe0ce3ab0d034e63669e426e294e4d # v1.2.2
+
+      - name: Login to GHCR
+        run: |
+          echo "${{ secrets.GITHUB_TOKEN }}" | oras login ghcr.io -u ${{ github.actor }} --password-stdin
+
+      - name: Mirror multi-platform tooling images
+        env:
+          SOURCE_IMAGE: ${{ matrix.images }}
+        run: |
+          if [[ "$SOURCE_IMAGE" == */* ]]; then
+            TARGET_IMAGE=$(echo "$SOURCE_IMAGE" | sed 's#^[^/]*/##')
+          else
+            TARGET_IMAGE="$SOURCE_IMAGE"
+          fi
+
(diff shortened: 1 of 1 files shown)


.github/workflows/mirror-tooling-images.yml
```diff
@@ -0,0 +1,59 @@
...
+            # Image with registry hostname (strip hostname)
+            TARGET_IMAGE=$(echo "$SOURCE_IMAGE" | sed 's#^[^/]*/##')
+          else
+            # Official library images (no registry hostname)
+            TARGET_IMAGE="$SOURCE_IMAGE"
+          fi
+
+          echo "Processing image: $SOURCE_IMAGE -> $GHCR_NAMESPACE/$TARGET_IMAGE"
+
+          docker pull "$SOURCE_IMAGE"
```
sozercan, line 57: can we prefix with mirror.gcr.io so it uses the google pull through cache? otherwise, we'll get throttled here

.github/workflows/mirror-tooling-images.yml
```diff
@@ -0,0 +1,59 @@
...
+    strategy:
+      matrix:
+        images:
+          - "debian:10-slim"
+          - "debian:11-slim"
+          - "debian:12-slim"
+          - "debian:stable-slim"
+          - "ubuntu:20.04"
+          - "ubuntu:22.04"
+          - "ubuntu:latest"
```
sozercan, line 32: is this one being used?

.github/workflows/mirror-tooling-images.yml
robert-cronin: good point, we wouldn't want the github action itself to get throttled. I'll update! presumably we can keep the mcr images as is and not worry about throttling?

.github/workflows/mirror-tooling-images.yml
robert-cronin: only in the workflows so I've removed this one

robert-cronin:
new run with changes: https://github.com/robert-cronin/copacetic/actions/runs/13692651586/job/38288503671

robert-cronin:
Split this out into this and #947 to avoid temporary resolution issues. This PR should be merged first to get the images into ghcr.io.

.github/workflows/mirror-tooling-images.yml
sozercan: yea i was debating that, let's keep it to be consistent and in case mcr introduces throttling later?

.github/workflows/mirror-tooling-images.yml
```diff
@@ -0,0 +1,58 @@
...
+            # Image with registry hostname (strip hostname)
+            TARGET_IMAGE=$(echo "$SOURCE_IMAGE" | sed 's#^[^/]*/##')
+          else
+            # Official library images (no registry hostname)
+            TARGET_IMAGE="$SOURCE_IMAGE"
+          fi
+
+          echo "Processing image: $SOURCE_IMAGE -> $GHCR_NAMESPACE/$TARGET_IMAGE"
+
+          docker pull "$SOURCE_IMAGE"
```
sozercan, line 56: is this going to work with multi platform images? i think we'll need oras, crane or containerd image store maybe?

.github/workflows/mirror-tooling-images.yml
```diff
@@ -0,0 +1,58 @@
...
+          - "mirror.gcr.io/debian:12-slim"
+          - "mirror.gcr.io/debian:stable-slim"
+          - "mirror.gcr.io/ubuntu:20.04"
+          - "mirror.gcr.io/ubuntu:22.04"
+          - "mcr.microsoft.com/azurelinux/base/core:3.0"
+          - "mcr.microsoft.com/cbl-mariner/base/core:2.0"
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
```
sozercan, line 37: pin to digest and add comment with version, you can see other actions as an example

.github/workflows/mirror-tooling-images.yml
robert-cronin: it seems the mcr images are not on gcr or I couldn't find them at least, do you know if they are?

.github/workflows/mirror-tooling-images.yml
robert-cronin: ahh, completely missed multi arch consideration. changed to using oras now, thanks!

.github/workflows/mirror-tooling-images.yml
robert-cronin: done!

robert-cronin:
new run with multi arch: https://github.com/robert-cronin/copacetic/actions/runs/13711818449

sozercan (approved):
lgtm, thanks!

PR #961, opened 2025-03-17
feat: add support for custom suffix tag

### LINKED ISSUE #923: [REQ] Add support for custom suffix tag

None

Instead of "-patched" as a default tag suffix, consider adding a custom suffix flag.

robert-cronin:
Adds a tag suffix to the patch command
Adds testing for the tag suffix logic
Adds extra testing to the patch command

Closes #923

### DIFF
--- a/pkg/patch/cmd.go
+++ b/pkg/patch/cmd.go
@@ -19,6 +19,7 @@ type patchArgs struct {
 	appImage      string
 	reportFile    string
 	patchedTag    string
+	suffix        string
 	workingFolder string
 	timeout       time.Duration
 	scanner       string
@@ -46,6 +47,7 @@ func NewPatchCmd() *cobra.Command {
 				ua.appImage,
 				ua.reportFile,
 				ua.patchedTag,
+				ua.suffix,
 				ua.workingFolder,
 				ua.scanner,
 				ua.format,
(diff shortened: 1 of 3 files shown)


ashnamehrotra (changes requested):

pkg/patch/cmd.go
```diff
@@ -58,6 +60,7 @@ func NewPatchCmd() *cobra.Command {
 	flags.StringVarP(&ua.appImage, "image", "i", "", "Application image name and tag to patch")
 	flags.StringVarP(&ua.reportFile, "report", "r", "", "Vulnerability report file path")
 	flags.StringVarP(&ua.patchedTag, "tag", "t", "", "Tag for the patched image")
+	flags.StringVarP(&ua.suffix, "tag-suffix", "", "", "Suffix for the patched image (if no explicit --tag provided)")
```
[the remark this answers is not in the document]
robert-cronin: I've added the default to the cli and changed the `resolvePatchedTag` logic to support this

pkg/patch/patch.go
```diff
@@ -337,6 +326,34 @@ func patchWithContext(ctx context.Context, ch chan error, image, reportFile, pat
...
+	}
+
+	// parse out any existing tag from the image ref
+	var baseTag string
+	if tagged, ok := imageRef.(reference.Tagged); ok {
+		baseTag = tagged.Tag()
+	}
+
+	// if suffix is empty, default to "patched"
+	if suffix == "" {
```
ashnamehrotra, line 343: I dont think it will reach this case since we have a default value?

pkg/patch/patch.go
```diff
@@ -337,6 +326,34 @@ func patchWithContext(ctx context.Context, ch chan error, image, reportFile, pat
...
+		suffix = "patched"
+	}
+
+	// if we have no original baseTag (the user’s image had no tag),
+	// then final name is just the suffix
+	if baseTag == "" {
+		return suffix, nil
+	}
```
ashnamehrotra, line 351: could this return in an error if the original image is just `test` and the patched image becomes `test:-patched`?

ashnamehrotra (changes requested):

pkg/patch/patch.go
robert-cronin: This part is in case the user passes `--tag-suffix=""` explicitly. I am not sure why one would want to do that but good to cover all edge cases, I guess. Please let me know if I can amend this logic in any way to be more consistent with copa standards.

ashnamehrotra: got it, that makes sense!

ashnamehrotra (approved):
LGTM, thanks!

PR #973, opened 2025-03-19
feat: invalidate tdnf cache when patching

### LINKED ISSUE #967: [REQ] invalidate tdnf cache when patching

None

if the tdnf repo is cached (due to image running `tdnf update` for example), copa may unable to pull in the repo to check for updates until the cache expires
copa should invalidate the cache to make sure to check for updates 

Use either `clean` or `upgrade --refresh`:

> clean: This command cleans up temporary files, data, and metadata. It takes the argument all.

> You can also run the upgrade command with the refresh option to update the cached metadata with the latest information from the repositories. 

https://vmware.github.io/photon/docs-v4/administration-guide/managing-packages-with-tdnf/standard-syntax-for-tndf/commands/

metadata expiry:

> The metadata_expire setting specifies the expiry time limit for the downloaded metadata in seconds. After the set limit expires, metadata is refreshed on the next action that requires them. The default value is 172800 seconds.

https://vmware.github.io/photon/docs-v5/administration-guide/managing-packages-with-tdnf/adding-a-new-repository/

i haven't verified if the same issue exists for yum, dnf, etc too but most likely

robert-cronin:
Adds cache cleaning commands to rpm manager

Closes #967

### DIFF
--- a/integration/fixtures/tdnf-test-img/Dockerfile
+++ b/integration/fixtures/tdnf-test-img/Dockerfile
@@ -1,0 +1,2 @@
+FROM mcr.microsoft.com/azure-watson/agent/agent_mariner:1.21.13.0@sha256:99641f3f0042ee4b2170b65878486f339e0db2f6efb4cc36eda942f68e283fe1
+RUN tdnf update -y && tdnf upgrade -y
--- a/integration/fixtures/test-images.json
+++ b/integration/fixtures/test-images.json
@@ -192,5 +192,13 @@
         "distro": "Mariner Distroless",
         "description": "Custom rpmmanifest files, no yum/tdnf/dnf/microdnf/rpm, custom openssl.cnf",
         "ignoreErrors": false
+    },
+    {
+        "image": "ghcr.io/project-copacetic/copacetic/test/tdnf",
+        "tag": "test",
+        "digest": "sha256:2920efc9c8447ad404cde391eaa6f61f39a2c2e7dffb678f241f36be6ede0f33",
+        "distro": "Mariner",
+        "description": "Valid rpm DB, tdnf, yum & rpm present, custom tdnf.conf",
+        "ignoreErrors": false
     }
 ]
--- a/pkg/pkgmgr/rpm.go
+++ b/pkg/pkgmgr/rpm.go
@@ -493,15 +493,15 @@ func (rm *rpmManager) installUpdates(ctx context.Context, updates unversioned.Up
 		if dnfTooling == "" {
 			dnfTooling = rm.rpmTools["dnf"]
 		}
-		checkUpdateTemplate := `sh -c 'if [ "$(%[1]s -q check-update | wc -l)" -ne 0 ]; then echo >> /updates.txt; fi'`
+		checkUpdateTemplate := `sh -c '%[1]s clean all && %[1]s makecache --refresh -y; if [ "$(%[1]s -q check-update | wc -l)" -ne 0 ]; then echo >> /updates.txt; fi'`
 		if !rm.checkForUpgrades(ctx, dnfTooling, checkUpdateTemplate) {
 			return nil, nil, fmt.Errorf("no patchable packages found")
 		}
 
-		const dnfInstallTemplate = `sh -c '%[1]s upgrade %[2]s -y && %[1]s clean all'`
+		const dnfInstallTemplate = `sh -c '%[1]s upgrade --refresh %[2]s -y && %[1]s clean all'`
 		installCmd = fmt.Sprintf(dnfInstallTemplate, dnfTooling, pkgs)
 	case rm.rpmTools["yum"] != "":
-		checkUpdateTemplate := `sh -c 'if [ "$(%[1]s -q check-update | wc -l)" -ne 0 ]; then echo >> /updates.txt; fi'`
+		checkUpdateTemplate := `sh -c '%[1]s clean all && %[1]s makecache fast; if [ "$(%[1]s -q check-update | wc -l)" -ne 0 ]; then echo >> /updates.txt; fi'`
 		if !rm.checkForUpgrades(ctx, rm.rpmTools["yum"], checkUpdateTemplate) {
 			return nil, nil, fmt.Errorf("no patchable packages found")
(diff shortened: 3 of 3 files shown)


ashnamehrotra:
@robert-cronin LGTM, is there an image we can test the before/after with?

robert-cronin:
> @robert-cronin LGTM, is there an image we can test the before/after with?


Great idea! in terms of validation, here are a couple of things I tried after seeing your comment (everything seems to be working okay):
- I ran copa on an image with tdnf and observed the following output:
```
#12 sh -c /usr/bin/tdnf clean all && /usr/bin/tdnf makecache --refresh -y; if [ "$(/usr/bin/tdnf -q check-update | wc -l)" -ne 0 ]; then echo >> /updates.txt; fi
#12 0.077 Loaded plugin: tdnfrepogpgcheck
#12 0.087 cleaning mariner-official-extras: metadata dbcache packages keys expire-cache
#12 0.087 cleaning mariner-official-base-source: metadata dbcache packages keys expire-cache
#12 0.087 cleaning mariner-official-base-debuginfo: metadata dbcache packages keys expire-cache
#12 0.087 cleaning mariner-official-base: metadata dbcache packages keys expire-cache
#12 0.087 cleaning mariner-official-microsoft: metadata dbcache packages keys expire-cache
#12 0.087 Done.
#12 0.095 Loaded plugin: tdnfrepogpgcheck
#12 0.102 Refreshing metadata for: 'CBL-Mariner Official Extras 2.0 x86_64'
#12 0.797 Refreshing metadata for: 'CBL-Mariner Official Base 2.0 x86_64'
#12 14.43 Refreshing metadata for: 'CBL-Mariner Official Microsoft 2.0 x86_64'
#12 17.70 Metadata cache created.
#12 DONE 17.9s

#13 sh -c /usr/bin/tdnf upgrade --refresh nghttp2 curl-libs expat-libs glibc libcap ncurses-libs openssl sqlite-libs openssl-libs zstd-libs curl expat krb5 libarchive libssh2 libtasn1 -y && /usr/bin/tdnf clean all
#13 0.077 Loaded plugin: tdnfrepogpgcheck
#13 0.089 Refreshing metadata for: 'CBL-Mariner Official Extras 2.0 x86_64'
#13 0.753 Refreshing metadata for: 'CBL-Mariner Official Base 2.0 x86_64'
...
```
- I just noticed there is a test case that targets rpm & tdnf in the integration tests, the job run of which can be found here: https://github.com/project-copacetic/copacetic/actions/runs/13980360640/job/39144205501?pr=973#step:12:59 (theres also ones that cover yum and microdnf too)

Please let me know if there is anything else I missed or anything else I can do to validate, thank you!

sozercan:
@robert-cronin can we create a test image to replicate this behavior and add an integration test?

robert-cronin:
Hi @sozercan do you have an existing image that exhibits the issue mentioned above? I think we can create one with metadata_expire=never but we'd have to wait for another vulnerability/fix to be reported and to not be included in the cache. Unless there is a way we can prevent upgrade/update from refreshing the cache?

robert-cronin:
> @robert-cronin can we create a test image to replicate this behavior and add an integration test?
> 
> something like (please validate this triggers the cache issue)
> 
> ```
> FROM mcr.microsoft.com/cbl-mariner/base/core:2.0@sha256:2a5d3461de4c082b1ced83a491c0d83b80221311dbee1b6f0a98271cefe57b00
> RUN tdnf update -y && tdnf upgrade -y
> ```
> 
> see the recent test image that @ashnamehrotra created for openssl and symlink bugs integration testing https://github.com/orgs/project-copacetic/packages/container/package/copacetic%2Ftest%2Fopenssl perhaps we can add to that image instead of creating a new one?
> 
> additionally, do we see same behavior with other package managers? can we handle as a follow up?

Hi @sozercan do you know of an existing image that exhibits the issue? I think we can create one with `metadata_expire=never` but we'd have to wait for another vulnerability/fix to be reported and to not be included in the cache already. Unless there is a way we can prevent upgrade/update from refreshing the cache before updating? So far I've tried: forging the lastrefresh timestamp to be less than 48 hours, setting the metadata_expire to never in the yum repos under `/etc/yum.repod.d/*.repo`, but the tdnf update/upgrade commands' cache refresh logic seem to be quite sticky

robert-cronin:
As for why this happens, the only difference I've seen in the problem image (i.e. `mcr.microsoft.com/azure-watson/agent/agent_mariner:1.21.13.0`) and others like base/core is that the problem image already has the cache there (even if its expired). So, if you add it in like this:

```
FROM mcr.microsoft.com/azure-watson/agent/agent_mariner:1.21.13.0@sha256:99641f3f0042ee4b2170b65878486f339e0db2f6efb4cc36eda942f68e283fe1 AS cache
RUN tdnf update -y && tdnf upgrade -y

# copy over the cache
FROM mcr.microsoft.com/cbl-mariner/base/core:2.0
COPY --from=cache /var/cache/tdnf /var/cache/tdnf
```

then try to run `tdnf update -y` in the base/core image, it will replicate the issue. This is odd because I would have thought since the cache is about 1 month old it would refresh automatically. This is probably just surprising to me due to my ignorance of how tdnf caching works.

integration/fixtures/tdnf-test-img/Dockerfile
```diff
@@ -0,0 +1,2 @@
+FROM mcr.microsoft.com/azure-watson/agent/agent_mariner:1.21.13.0@sha256:99641f3f0042ee4b2170b65878486f339e0db2f6efb4cc36eda942f68e283fe1
+RUN tdnf update -y && tdnf upgrade -y
```
ashnamehrotra, line 2: nit newline

ashnamehrotra (approved):

ashnamehrotra:
@robert-cronin this fix and test lgtm, did we confirm if this is an issue with other package managers?

robert-cronin:
> @robert-cronin this fix and test lgtm, did we confirm if this is an issue with other package managers?

No I haven't checked any other package managers for this particular issue yet. I'll create the issues for it!

PR #978, opened 2025-03-21
fix: rpm fails if mirror tooling image not present

### LINKED ISSUE #976: [BUG] copa fails if mirrored tooling images not present
### Version of copa

_No response_

### Expected Behavior

All tooling images should be present. If not, they should fall back to using the original instead of failing to resolve a missing image at `ghcr.io/project-copacetic/copacetic`

### Actual Behavior

Recently discovered this behavior while trying to patch an old version of `mcr.microsoft.com/azure-watson/agent/agent_mariner`

```
❯ ./dist/linux_amd64/release/copa patch --image mcr.microsoft.com/azure-watson/agent/agent_mariner:1.17.100.00-230720  -s trivy -r /tmp/tdnf-report.json -a buildx://copademo
[+] Building 1.0s (2/3)                                                                                                                                                                                                
 => resolve image config for docker-image://mcr.microsoft.com/azure-watson/agent/agent_mariner:1.17.100.00-230720                                                                                                 0.2s
 => ERROR docker-image://ghcr.io/project-copacetic/copacetic/cbl-mariner/base/core:1.0                                                                                                                            0.9s
 => => resolve ghcr.io/project-copacetic/copacetic/cbl-mariner/base/core:1.0                                                                                                                                      0.9s
Error: failed to load cache key: ghcr.io/project-copacetic/copacetic/cbl-mariner/base/core:1.0: not found
```

### Steps To Reproduce

1. just run `copa patch --image mcr.microsoft.com/azure-watson/agent/agent_mariner:1.17.100.00-230720`

robert-cronin:
Add fallback mechanism for when tooling image is not present for rpm-based images

Closes #976


robert-cronin:
I believe the specific image (a 1.0 tag from cbl-mariner/core) in question is not one that is expected to be present in the tooling images cache, and if so, it probably not an image we will see pulled often as it is fairly dated.

sozercan:
@robert-cronin do we want to add core:1.0 to the ghcr cache too?

sozercan (approved):
thank you! lgtm. added a few suggestions that we can handle as follow ups

robert-cronin:
> thank you! lgtm. added a few suggestions that we can handle as follow ups

Excellent thank you for your review! 

> @robert-cronin do we want to add core:1.0 to the ghcr cache too?
> 
> do we want to add the same mechanism for other package managers in case there's a new version but we are not caching it yet?

I've added the following based on suggestions above:

- https://github.com/project-copacetic/copacetic/issues/980
- https://github.com/project-copacetic/copacetic/pull/979

Also, I just noticed there was a call to `toolingImage.Validate` i forgot to clean up so that is latest push you'll notice in this PR

ashnamehrotra (approved):

PR #1004, opened 2025-04-01
Add push images on patch flag

### LINKED ISSUE #293: [REQ] Support pushing to registry directly

robert-cronin:
Adds --push command, defaults to false
Adds e2e test with a local registry

Closes #293


test/e2e/push_test.go
```diff
@@ -0,0 +1,154 @@
...
+				]
+			}
+		]
+	}`
+	err = os.WriteFile(reportFile, []byte(reportContent), 0o600)
+	require.NoError(t, err, "failed to create sample report file")
+
+	// run copa patch with push flag
+	targetImage := "localhost:5000/alpine:patched"
+	patchCmd := exec.Command("copa", "patch",
```
ashnamehrotra, line 90: I think test is failing due to this since we did not provide copa path

ashnamehrotra (changes requested):

ashnamehrotra (approved):
LGTM after rebase and small formatting edits

PR #1006, opened 2025-04-01
fix: lint fixes for ci

ashnamehrotra:

Closes #<_issue_ID_>


sozercan (approved):

robert-cronin:
Do we also need to update the assertion in the patch test? I think its failing because it's expecting a certain character casing

PR #1009, opened 2025-04-03
feat: add multi arch command changes

### LINKED ISSUE #988: Multi-Arch - patch Command Changes
### `patch` Command Changes:

- Modify the patch command flow (`cmd/patch/patch.go`) to:
  1. Detect if the input reference is a multi-arch manifest using the utility built in #989 
  2. Retrieve sub-manifests and store them in a slice/struct describing each (OS, Arch, Digest).
  3. When multi-arch is detected, loop over each sub-manifest to call `Patch`; otherwise just call `Patch` as before.

Dependent on #989

robert-cronin:
- added QemuAvailable() + mapGoArch() helpers to detect binfmt/qemu support for cross-arch builds.
- new helpers for multiarch work:
  - archDigest, archTag(), dockerImageDigest(), createManifestCLI() and normalizeConfigForPlatform()
- split the monolithic patchWithContext() into:
  - patchSingleArchImage() - one arch at a time (returns *types.PatchResult).
  - patchMultiArchImage() - fan-out, run builds in parallel, then stitch a manifest list.
- added new types:
  - PatchPlatform: OCI platform + report path
  - PatchResult: original image, patched ref, digest
- added argument validation: can't pass both report-file and report-dir

Closes #988

### DIFF
--- a/pkg/buildkit/buildkit.go
+++ b/pkg/buildkit/buildkit.go
@@ -7,6 +7,8 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"os/exec"
+	"strings"
 
 	"github.com/containerd/platforms"
 	"github.com/moby/buildkit/client/llb"
@@ -15,6 +17,7 @@ import (
 	ispec "github.com/opencontainers/image-spec/specs-go/v1"
 
 	"github.com/project-copacetic/copacetic/pkg/report"
+	"github.com/project-copacetic/copacetic/pkg/types"
 	log "github.com/sirupsen/logrus"
 
 	"github.com/google/go-containerregistry/pkg/name"
@@ -39,6 +42,13 @@ type Opts struct {
 }
 
 const linux = "linux"
+
+// for testing.
+var (
+	readDir  = os.ReadDir
+	readFile = os.ReadFile
+	lookPath = exec.LookPath
+)
 
 func InitializeBuildkitConfig(ctx context.Context, c gwclient.Client, userImage string) (*Config, error) {
 	// Initialize buildkit config for the target image
@@ -90,8 +100,8 @@ func InitializeBuildkitConfig(ctx context.Context, c gwclient.Client, userImage 
 	return &config, nil
 }
 
-func DiscoverPlatformsFromReport(reportDir, scanner string) ([]ispec.Platform, error) {
-	var platforms []ispec.Platform
+func DiscoverPlatformsFromReport(reportDir, scanner string) ([]types.PatchPlatform, error) {
+	var platforms []types.PatchPlatform
 
 	reportNames, err := os.ReadDir(reportDir)
 	if err != nil {
@@ -99,7 +109,11 @@ func DiscoverPlatformsFromReport(reportDir, scanner string) ([]ispec.Platform, e
 	}
 
 	for _, file := range reportNames {
-		report, err := report.TryParseScanReport(reportDir+"/"+file.Name(), scanner)
+		filePath := reportDir + "/" + file.Name()
+		if file.IsDir() {
+			continue
+		}
+		report, err := report.TryParseScanReport(filePath, scanner)
 		if err != nil {
 			return nil, fmt.Errorf("error parsing report %w", err)
 		}
@@ -109,9 +123,12 @@ func DiscoverPlatformsFromReport(reportDir, scanner string) ([]ispec.Platform, e
 			continue
 		}
 
-		platform := ispec.Platform{
-			OS:           linux,
-			Architecture: report.Metadata.Config.Arch,
+		platform := types.PatchPlatform{
+			Platform: ispec.Platform{
+				OS:           linux,
+				Architecture: report.Metadata.Config.Arch,
+			},
+			ReportFile: filePath,
 		}
 		platforms = append(platforms, platform)
 	}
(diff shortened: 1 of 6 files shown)


robert-cronin:
This is a convenience script I am using to generate a reports directory, just going to leave this here:

```
#!/usr/bin/env bash
set -euo pipefail

IMAGE="alpine:3.19"
REPORT_DIR="/tmp/reports"
rm -rf "$REPORT_DIR" && mkdir -p "$REPORT_DIR"

PLATFORMS=(
  linux/amd64
  linux/arm/v5
  linux/arm/v6
  linux/arm/v7
  linux/arm64
  linux/386
  linux/mips64le
  linux/ppc64le
  linux/riscv64
  linux/s390x
)

for platform in "${PLATFORMS[@]}"; do
  suffix=${platform//\//-}
  out="$REPORT_DIR/report-${suffix}.json"

  trivy image \
    --platform "$platform" \
    --image-src remote \
    --pkg-types os \
    --ignore-unfixed \
    --db-repository public.ecr.aws/aquasecurity/trivy-db:2 \
    --format json \
    -o "$out" \
    "$IMAGE" \
done
```

robert-cronin:
This is a convenience script I am using to generate a reports directory, just going to leave this here:

```
#!/usr/bin/env bash
set -euo pipefail

IMAGE="nginx:1.21.6"
REPORT_DIR="/tmp/reports"
rm -rf "$REPORT_DIR" && mkdir -p "$REPORT_DIR"

PLATFORMS=(
  linux/amd64
  linux/arm/v5
  linux/arm64
  linux/386
  linux/mips64le
  linux/ppc64le
  linux/s390x
)

for platform in "${PLATFORMS[@]}"; do
  suffix=${platform//\//-}
  out="$REPORT_DIR/report-${suffix}.json"

  trivy image \
    --platform "$platform" \
    --image-src remote \
    --pkg-types os \
    --ignore-unfixed \
    --format json \
    -o "$out" \
    "$IMAGE"
done
```

pkg/patch/patch.go
```diff
@@ -87,30 +150,107 @@ func removeIfNotDebug(workingFolder string) {
...
+			return fmt.Errorf("report file %s is a directory, please provide a file", reportFile)
+		}
+		log.Debugf("Using report file: %s", reportFile)
+		platform := types.PatchPlatform{
+			Platform: platforms.Normalize(platforms.DefaultSpec()),
+		}
+		if platform.OS != "linux" {
+			platform.OS = "linux"
+		}
+		_, err = patchSingleArchImage(ctx, ch, image, reportFile, patchedTag, suffix, workingFolder, scanner, format, output, platform, ignoreError, push, bkOpts, false)
```
ashnamehrotra, line 190: should we save and output the PatchResult over here?

pkg/patch/patch.go
```diff
@@ -213,28 +353,30 @@ func patchWithContext(ctx context.Context, ch chan error, image, reportFile, rep
...
 		case strings.Contains(solveOpt.SourcePolicy.Rules[0].Updates.Identifier, "alma"):
 			err = errors.New("AlmaLinux is not supported via source policies due to BusyBox not being in the AlmaLinux repos\n" +
 				"Please use a different RPM-based image")
-			return err
+			return nil, err
 		}
 	}
 
+	// Create a channel to receive the patched image digest
+	patchedImageDigest := make(chan string, 1)
```
ashnamehrotra, line 379: why do we need a channel here?

pkg/buildkit/buildkit.go
```diff
@@ -334,3 +346,101 @@ func WithFileString(s *llb.State, path, contents string) llb.State {
 func WithFileBytes(s *llb.State, path string, contents []byte) llb.State {
 	return s.File(llb.Mkfile(path, 0o600, contents))
 }
+
+func QemuAvailable(p *types.PatchPlatform) bool {
+	if p == nil {
```
ashnamehrotra, line 351: do you think this is a func we could add test for?

pkg/patch/patch.go
```diff
@@ -50,6 +54,65 @@ var (
...
+	if variant != "" {
+		return fmt.Sprintf("%s-%s-%s", base, arch, variant)
+	}
+	return fmt.Sprintf("%s-%s", base, arch)
+}
+
+// dockerImageDigest returns the image digest for a given image name.
+// It uses the "docker image inspect" command to get the digest.
+func dockerImageDigest(img string) (string, error) {
+	out, err := exec.Command("docker", "image", "inspect", "--format={{index .Id}}", img).Output()
```
ashnamehrotra, line 74: we might have already discussed this, but was there a library we could use for this?

pkg/buildkit/buildkit.go
robert-cronin: I've added some tests for this with mocking for the os apis

pkg/patch/patch.go
robert-cronin: This channel ferries the digest from the go routine out into the larger context to avoid races, but I've just realised since we use `eg.Wait`, there won't be any race, so I've changed it to be just a regular string var

pkg/patch/patch.go
robert-cronin: Added some logging for the result here

pkg/patch/patch.go
robert-cronin: I think we did something similar for local media type in the oci mediatype PR, I've changed it to use that library. Happy to use another library if there is one that can do this better!

ashnamehrotra (approved):

PR #1017, opened 2025-04-15
feat: Discover Platforms Utility

### LINKED ISSUE #989: Multi-Arch - Discover Platforms Utility
### Manifest Utility
- Add flag: `--report-dir`
- Implement a helper function (e.g. `func DiscoverPlatforms(imageRef, reportDir string) ([]Platform, error)`)  to:
  - Inspect the image reference using `go-containerregistry` or `BuildKit` and **from the contents of the report** discover a list of sub-manifest digests and platform info (if there are multiple reports claiming to describe the same manifest, either fail or take the first one); then
  - Read the reportDir and build a list of patchable platforms; then
  - Return the intersection of these 

Note: may need to add fields to `Platform` type

### Handle failures for one or more archs
- Add a flag to expose control of this behavior to the user (e.g. `--platform-specific-errors=[skip|warn|fail]`)

ashnamehrotra:

Closes #989

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -49,6 +49,7 @@ require (
 	github.com/containerd/errdefs v1.0.0 // indirect
 	github.com/containerd/errdefs/pkg v0.3.0 // indirect
 	github.com/containerd/log v0.1.0 // indirect
+	github.com/containerd/stargz-snapshotter/estargz v0.16.3 // indirect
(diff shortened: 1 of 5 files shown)


robert-cronin (approved):
This is working okay on my local! tested with nginx and amd64+arm64. I think this should be good to merge after we solve the linter issues

PR #1053, opened 2025-05-13
chore: Fix some issues with docker and buildx connections

### LINKED ISSUE #543: [REQ] Update buildx integration to use new `docker buildx dial-stdio`

In 3dec6b76d0f35442e5733e8f530c7fa8fd2df450 we added support for connecting to a buildx builder, with the limitation that it must either be a docker-container backed builder OR the default builder.
It also re-implements some things that buildx does internally for figuring what builder to connect to and making sure it is started up.
Some of this is relying on internal config files under ~/.docker/buildx` that aren't really documented and are just part of how buildx configures itself.

With buildx v0.13 [we added support for a new command `docker buildx dial-stdio`](https://github.com/docker/buildx/pull/2112).
This new command uses the stdin and stdout streams of the command as the transport for the buildkit client.
With this new command we no longer need to peek into the internal buildx config or manually bootstrap the builder instance.
We also get access to all the drivers that buildx supports.

I propose we remove the old logic and switch to this new command, with the caveat that it requires buildx 0.13.
Example usage: https://github.com/Azure/dalec/blob/3593addf5b0f57cc5ca6fc21705562fa545e93e8/cmd/localdev/main.go#L150-L170

cpuguy83:
1. Get the docker addr from "docker context ls" when `DOCKER_HOST` is not set
2. Fix issue when buildx metadata specifies a docker context name rather than a docker connection string
3. Add support for `buildx dial-stdio` which will proxy connections to buildkit for us rather than manually setting up the connection.

This makes it so `copa` works with Docker Desktop out of the box (no extra settings need to be changed).
It also removes the limitation on the type of buildx instances it can connect to when the installed buildx supports the "dial-stdio" command (which has been around for awhile now).

Closes #543

### DIFF
--- a/pkg/buildkit/connhelpers/buildx.go
+++ b/pkg/buildkit/connhelpers/buildx.go
@@ -1,17 +1,20 @@
 package connhelpers
 
 import (
+	"bufio"
 	"bytes"
 	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"math/rand"
 	"net"
 	"net/url"
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strings"
 
 	"github.com/cpuguy83/dockercfg"
 	"github.com/cpuguy83/go-docker"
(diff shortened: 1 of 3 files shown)


ashnamehrotra (approved):
LGTM

Issue #1010, opened 2025-04-04
[REQ] Improve error logs


ashnamehrotra:
### What kind of request is this?

None

### What is your request or suggestion?

Improve error logs for the following cases

- EOSL
- Copa throttling 
- for "no upgradable packages" make sure log is consistent amongst package managers

### Are you willing to submit PRs to contribute to this feature request?

- [ ] Yes, I am willing to implement it.

amanycodes:
@ashnamehrotra Can I take this up?

amanycodes:
@ashnamehrotra what exactly do we mean by copa "throttling" here.🤔
I'm not sure I quite understand it yet.

ashnamehrotra:
@amanycodes this should have been fixed via #937, so we can skip that log now!

PR #1069, opened 2025-05-20
feat: Improve EOL Logging

amanycodes:
Added checks for whether a container Image is `EOL` using the [endoflife.date](https://endoflife.date/) [API](https://endoflife.date/docs/api/v1/). Now the user will get a `WARN` level log in the console if the image is `EOL` without blocking the patching. Eg:
 `WARN[0005] The operating system ubuntu 18.04 appears to be End-Of-Support-Life. (EOL date: 2023-05-31) Patching may fail, be incomplete, or use archived repositories. Consider upgrading the base image.`

I faced an issue while developing this where `EOL` `WARN` level logs did not appear in the console when I ran `copa` without `--debug` flag, even though the logging conditions were satisfied. I tried debugging this problem by checking `WARN` level logs for other logs like `update all` mode, etc. and they seemed to work. Please let me know where am I going wrong :(

I have added basic tests for now. Once the logic is approved, I'll improve the tests!

Partially closes: #1010


ashnamehrotra:
@amanycodes thanks for the PR - the logic looks good to me! that is the expected behavior with/without the --debug flag, it is a little flakey with logging sometimes due to cache.

ashnamehrotra:
@amanycodes sounds good, thanks!

amanycodes:
@ashnamehrotra there are just helper function tests in the pkg/patch/patch_test.go (how do I fix the codegov coverage :thinking: ) so I tried adding integration tests which are constantly failing. tried bypassing the trivy scans and even checked `"ignoreErrors": true` but can't seem to find a workaround :(
Please let me know where am I going wrong and is integration test really needed or can we proceed without it for now?

PR #1070, opened 2025-05-22
feat: add manual rule hash to hash patching

### LINKED ISSUE #1033: [REQ] Manual "hash to hash" patch rules for images without supported package managers

New feature

Copas current patch flow depends on the target image exposing a package manager (apk, dpkg, rpm) so it can:

1. locate an installed version (i.e. fixed version mapping) and
2. apply an update with dependency/backwards compatible safety

That leaves a gap for images that do not ship a conventional package manager (e.g. scratch images, tiny statically linked distroless builds, language specific buildpacks) or for scenarios where operators already know exactly which artifact hash remediates a CVE and simply want to swap it in

e.g. there might be a CVE that shows up for a non-packaged file when using rekor:

```
trivy image --sbom-sources rekor <image>
```

robert-cronin:
Adds manual rule patching
Adds to docs to explain new behaviour

Closes #1033

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -354,3 +354,42 @@ jobs:
         run: |
           set -eu -o pipefail
           go test -v ./test/e2e/push --addr="docker://" --copa="$(pwd)/copa"
+
+  test-manual-rule:
+    needs: build
+    name: Test manual rule
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    permissions: read-all
+    steps:
+      - name: Harden Runner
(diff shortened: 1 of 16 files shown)


sozercan:
Let's hold on this until we can find a scenario to support this

robert-cronin:
> Let's hold on merging this until we can find a scenario to support this

Good idea, I am not sure if they're the right kind of scenarios, but I added some I could think of here: https://github.com/project-copacetic/copacetic/issues/1033#issuecomment-2917933571

robert-cronin:
Closing this for now in lieu of the app-level patching. We could always reopen if someone needs it down the line or theres a specific use case in mind.

PR #1079, opened 2025-05-29
feat: Add native v1alpha1 doc support

### LINKED ISSUE #857: [REQ] Support injesting copacetic's native report format

Given that [copacetic has defined it's own report format ](https://project-copacetic.github.io/copacetic/website/scanner-plugins),
when I have a report generated using that provided format, I would like it if copa could injest that report without needing to execute a scanner plugin.

Currenty, copa defaults to expecting a trivy report, and there does not seem to be any way to communicate to it that it should be looking for a report in it's own reporting structure. To work around that limitation I have to create a dummy scanner plugin using just a bash script of `cat $1` so that it will accept that report format.

amanycodes:
Added native v1alpha1 report support to Copa.
Now the users can specify report which is in the native `v1alpha1.UpdateManifest` JSON format using the `--scanner (or -s) native` flag
eg.
`copa patch -i <image> -s native -r ./v1alpha1-supported.json`

Closes #857

### DIFF
--- a/pkg/report/report.go
+++ b/pkg/report/report.go
@@ -3,6 +3,7 @@ package report
 import (
 	"encoding/json"
 	"fmt"
+	"os"
 	"os/exec"
 
 	"github.com/project-copacetic/copacetic/pkg/types/unversioned"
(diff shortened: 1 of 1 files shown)


ashnamehrotra:
@amanycodes 
I am wondering if we can change this to remove the need to provide the scanner completely. That way we don't need to plug-in the scanner or "native" as the argument, and can do something like "--custom-scanner"?

I was thinking since we are checking for `v1alpha1.UpdateManifest` format anyways once we unmarshal into `var m`, we can assume correct format if user decides to use custom scanner and instead of executing the scanner we can just read the file?

robert-cronin:
> @amanycodes I am wondering if we can change this to remove the need to provide the scanner completely. That way we don't need to plug-in the scanner or "native" as the argument, and can do something like "--custom-scanner"?
> 
> I was thinking since we are checking for `v1alpha1.UpdateManifest` format anyways once we unmarshal into `var m`, we can assume correct format if user decides to use custom scanner and instead of executing the scanner we can just read the file?
> 
> cc @sozercan @robert-cronin

I'd probably lean towards being explicit and having the `--scanner native` flag as it indicates to users they're providing a file already in copas v1alpha1 format, but i see the point about removing the cognitive overhead for users.

ashnamehrotra:
@amanycodes tested this and LGTM! can we also add this to the current documentation?

amanycodes:
Sure! updating the current documentation as well (in a different PR?).

ashnamehrotra:
> Sure! updating the current documentation as well (in a different PR?).

we can do it in this PR

ashnamehrotra:
@amanycodes  can we remove it from v0.10.0 docs? it will be part of v0.11.0 release

amanycodes:
@sozercan will this also be in v0.11 release?

sozercan:
@amanycodes yes, everything merged in to main before we cut the tag will be

Issue #992, opened 2025-03-27
Multi-Arch - Enhancements


robert-cronin:
Once the multi-arch patching logic is done, we might be able to improve UX by outputting a table summarizing which platforms were scanned, which were patched, and which were skipped so the user has a unified view.

amanycodes:
@robert-cronin I think the patching logic is mostly done (except the plugin system). Can I pick this up? Will start out with understanding the multi-arch changes and will raise a PR soon!

robert-cronin:
> [@robert-cronin](https://github.com/robert-cronin) I think the patching logic is mostly done (except the plugin system). Can I pick this up? Will start out with understanding the multi-arch changes and will raise a PR soon!

Hey @amanycodes, yes feel free! currently we are just logging out a simple info message here as each platform is patched: https://github.com/project-copacetic/copacetic/blob/main/pkg/patch/patch.go#L818 it would be great to see this instead as an aggregated summary of all results after they've all finished

PR #1098, opened 2025-06-06
feat: Add Multi Arch Enhancements

amanycodes:
Added table summary for Multi-Arch Patching with Platform, Status, Ref, Error as fields. Wanted feedback on how can I test this code. I'm confused on how to simulate multi-arch patching locally using docker and trivy to test the credibility and making sure the logs are as expected.
Please review the code and let me know this is how we wanted the logs to look. Thanks!

closes: #992

### DIFF
--- a/pkg/buildkit/buildkit.go
+++ b/pkg/buildkit/buildkit.go
@@ -209,6 +209,15 @@ func DiscoverPlatformsFromReference(manifestRef string) ([]types.PatchPlatform, 
 	return nil, nil
 }
 
+func PlatformKey(pl ispec.Platform) string {
+	// if platform is present in list from reference and report, then we should patch that platform
+	key := pl.OS + "/" + pl.Architecture
+	if pl.Variant != "" {
+		key += "/" + pl.Variant
+	}
+	return key
+}
+
 func DiscoverPlatforms(manifestRef, reportDir, scanner string) ([]types.PatchPlatform, error) {
 	var platforms []types.PatchPlatform
 
@@ -228,18 +237,13 @@ func DiscoverPlatforms(manifestRef, reportDir, scanner string) ([]types.PatchPla
 		}
 		log.WithField("platforms", p2).Debug("Discovered platforms from report")
 
-		// if platform is present in list from reference and report, then we should patch that platform
-		key := func(pl ispec.Platform) string {
-			return pl.OS + "/" + pl.Architecture + "/" + pl.Variant
-		}
-
 		reportSet := make(map[string]string, len(p2))
 		for _, pl := range p2 {
-			reportSet[key(pl.Platform)] = pl.ReportFile
+			reportSet[PlatformKey(pl.Platform)] = pl.ReportFile
 		}
 
 		for _, pl := range p {
-			if rp, ok := reportSet[key(pl.Platform)]; ok {
+			if rp, ok := reportSet[PlatformKey(pl.Platform)]; ok {
 				pl.ReportFile = rp
 				platforms = append(platforms, pl)
 			}
--- a/pkg/buildkit/buildkit.go
+++ b/pkg/buildkit/buildkit.go
@@ -209,6 +209,7 @@ func DiscoverPlatformsFromReference(manifestRef string) ([]types.PatchPlatform, 
 	return nil, nil
 }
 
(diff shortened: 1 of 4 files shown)


amanycodes:
The improved UX logs are working, i tested the multi-arch patching by building alpine:3.17.9 for amd64, arm64 and arm/v7 even though it ran just for the amd64 the `patchMultiArch` method was triggered and I got the tabular logs:
image

Opening this for review and ready for feedback!
cc: @robert-cronin

pkg/patch/patch.go
```diff
@@ -847,7 +867,26 @@ func patchMultiPlatformImage(
 		}
 	}
 
-	log.Infof("Multi-platform image patched with tag %s", patchedImageName.String())
+	var b strings.Builder
+	w := tabwriter.NewWriter(&b, 0, 0, 2, ' ', 0)
+	fmt.Fprintln(w, "PLATFORM\tSTATUS\tREFERENCE\tERROR")
+
+	for _, p := range platforms {
+		platformKey := p.OS + "/" + p.Architecture
```
robert-cronin: it looks like this is duplicated a few times, maybe we can pull it out into a shared function? There is an existing key in https://github.com/project-copacetic/copacetic/blob/main/pkg/buildkit/buildkit.go#L232. I like the conditional check for Variant. maybe we can combine these and place the function in buildkit.go and then import it here?

pkg/patch/patch_test.go
```diff
@@ -579,3 +580,68 @@ func TestNormalizeConfigForPlatform(t *testing.T) {
...
+		},
+		"linux/arm64": {
+			Platform: "linux/arm64",
+			Status:   "Error",
+			Ref:      "",
+			Error:    "emulation is not enabled for platform linux/arm64",
+		},
+		"linux/arm/v7": {
+			Platform: "linux/arm/v7",
+			Status:   "Skipped",
```
robert-cronin: Should this be "Ignored" instead?

robert-cronin (commented):
Nice work! I think that will make things much clearer for folks looking for a quick check. Just a couple of comments

pkg/patch/patch_test.go
amanycodes: Oh, yeah. missed this while resolving conflicts. will update it!

amanycodes:
@robert-cronin Thanks for the review! Please let me know if it looks fine now.

pkg/patch/patch.go
```diff
@@ -789,14 +784,36 @@ func patchMultiPlatformImage(
...
 
-			mu.Lock()
 			patchResults = append(patchResults, *res)
+			summaryMap[platformKey] = &types.MultiArchSummary{
+				Platform: platformKey,
+				Status:   "Patched",
+				Ref:      res.PatchedRef.String(),
+				Error:    "",
+			}
 			mu.Unlock()
```
robert-cronin: can we change this `mu.Unlock()` to defer until the function exits? I think we might run into trouble if the function returns before we unlock. We can place it next to the lock like so:
```
mu.Lock()
defer mu.Unlock()
```

pkg/patch/patch.go
```diff
@@ -789,14 +784,36 @@ func patchMultiPlatformImage(
...
+				if ignoreError {
+					status = "Ignored"
+				}
+				summaryMap[platformKey] = &types.MultiArchSummary{
+					Platform: platformKey,
+					Status:   status,
+					Ref:      "",
+					Error:    err.Error(),
+				}
+				return nil
```
robert-cronin, line 803: I think we should return the error if ignoreError is not true, this way we can fast fail up the chain

pkg/buildkit/buildkit.go
```diff
@@ -209,6 +209,15 @@ func DiscoverPlatformsFromReference(manifestRef string) ([]types.PatchPlatform,
 	return nil, nil
 }
 
+func PlatformKey(pl ispec.Platform) string {
```
robert-cronin, line 213: it looks like there is a linting error here for the size of pl. Were ignore this linting rule in patch.go so I think you can do the same here with the comment: `//nolint:gocritic`

amanycodes:
@robert-cronin Can I have your eyes on this? Thanks!

PR #1106, opened 2025-06-11
feat: .NET app-level patching

ashnamehrotra:

Closes #1106


ashnamehrotra:
@robert-cronin 

I pushed a `dot-net-test` image dir which patches dotnet vulns when testing.

To patch donet generally
- dotnet add package -v version
- dotnet restore (optional, dotnet add does by default)
- dotnet publish -c Release -o output (detecting this from filestructure, but structure could change)

However, there are some caveats
- I am not sure if the sdk will be provided in all images (in this case, we could mount it)
- In order to patch we also have to assume the project file is within the image, so it may be useful for a very specific case that I am not sure we want to maintain? Open to feedback or trying this for another language if we want instead.
- This patches lang vulns but will not patch SDK internal components ex/Microsoft.Build.Tasks.Core.. since these cannot be resolved without updating base image. This is why we will not be able to patch images such as `mcr.microsoft.com/dotnet/aspnet:6.0`

robert-cronin:
> @robert-cronin
> 
> I pushed a `dot-net-test` image dir which patches dotnet vulns when testing.
> 
> To patch donet generally
> 
> * dotnet add package -v version
> * dotnet restore (optional, dotnet add does by default)
> * dotnet publish -c Release -o output (detecting this from filestructure, but structure could change)
> 
> However, there are some caveats
> 
> * I am not sure if the sdk will be provided in all images (in this case, we could mount it)
> * In order to patch we also have to assume the project file is within the image, so it may be useful for a very specific case that I am not sure we want to maintain? Open to feedback or trying this for another language if we want instead.
> * This patches lang vulns but will not patch SDK internal components ex/Microsoft.Build.Tasks.Core.. since these cannot be resolved without updating base image. This is why we will not be able to patch images such as `mcr.microsoft.com/dotnet/aspnet:6.0`

I agree with all of these points. I guess we can use `mcr.microsoft.com/dotnet/sdk:<ver>` to mount the sdk, does the project file contain the necessary info to determine correct version of sdk to use?

Perhaps we can just scope it down to images that contain the necessary source files to rebuild and then note the supported use case in the docs.

.github/workflows/build.yml
```diff
@@ -813,3 +813,46 @@ jobs:
...
+        uses: actions/download-artifact@018cc2cf5baa6db3ef3c5f8a56943fffe632ef53 # v6.0.0
+        with:
+          name: copa_edge_linux_amd64.tar.gz
+      - name: Extract copa
+        shell: bash
+        run: |
+          tar xzf copa_edge_linux_amd64.tar.gz
+          ./copa --version
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@29109295f81e9208d7d86ff1c6c12d2833863392 # v3.6.0
```
robert-cronin: should this be v3.7?

pkg/langmgr/dotnet.go
```diff
@@ -0,0 +1,776 @@
...
+)
+
+type dotnetManager struct {
+	config        *buildkit.Config
+	workingFolder string
+}
+
+// isValidDotnetVersion checks if a version string is a valid semantic version.
+func isValidDotnetVersion(v string) bool {
+	_, err := semver.NewVersion(v)
```
robert-cronin: I think some dotnet packages might have `Major.Minor.Build.Revision` but not sure how common they are.

pkg/langmgr/dotnet.go
```diff
@@ -0,0 +1,776 @@
...
+			echo "WARN: $dll not found in runtime image, skipping"
+		fi
+	fi
+done
+
+# Copy native dependencies if they exist (runtimes/ folder)
+if [ -d "/output/runtimes" ]; then
+	echo "Copying native dependencies from runtimes/ folder"
+	if [ -d "$APP_DIR/runtimes" ]; then
+		cp -rf /output/runtimes/* "$APP_DIR/runtimes/"
```
robert-cronin, line 413: Is there a possibility that these dependencies not match the container architecture?

pkg/langmgr/dotnet.go
```diff
@@ -0,0 +1,776 @@
...
+		oldVersionEscaped := strings.ReplaceAll(oldVersion, ".", "\\.")
+
+		script.WriteString(fmt.Sprintf(`# Update %s from %s to %s
+echo "Updating %s: %s -> %s"
+echo "Before update:"
+grep -c "%s/%s" "$DEPS_FILE" || echo "Pattern not found"
+`, packageName, oldVersion, newVersion, packageName, oldVersion, newVersion, packageName, oldVersionEscaped))
+
+		// Use a temp file approach for safer sed operations
+		script.WriteString(fmt.Sprintf(`sed 's|"%s/%s"|"%s/%s"|g' "$DEPS_FILE" > "${DEPS_FILE}.tmp1" && mv "${DEPS_FILE}.tmp1" "$DEPS_FILE"
```
robert-cronin: Will this work for minified deps.json as well?

pkg/langmgr/dotnet.go
ashnamehrotra: yes, added this to the check!

pkg/langmgr/dotnet.go
ashnamehrotra: good point! I modified the sdk tooling image to be of the same architecture as the image we are patching

pkg/langmgr/dotnet.go
ashnamehrotra: I think it would still parse the same regardless of the whitespaces?

PR #1121, opened 2025-06-17
fix: preseve all  platforms in manifest

### LINKED ISSUE #1112: copy over windows digests (don't skip over them) in multi arch patching
since copa cannot patch windows containers, we need to keep them as is. this means we need to keep the existing windows container digests copied over to the new patched container index. this only applies to index and manifests list that includes non-host process windows containers, and it is not applicable to host-process containers.

this is a must before we can release multi platform functionality as this might lead to existing containers breaking.

ashnamehrotra:
Preserves digests of unpatched images (including Windows). I can modify this for just windows if we want, but thought it would be good to keep all digests in the resulting manifest.

Closes #1112


integration/multiarch/patch_test.go
```diff
@@ -126,6 +126,115 @@ func TestPatch(t *testing.T) {
 	}
 }
 
+// Tests patching only some architectures while preserving others.
+func TestPatchPartialArchitectures(t *testing.T) {
+	// Test image with multiple platforms including Windows
+	originalImage := "mcr.microsoft.com/oss/kubernetes-csi/secrets-store/driver"
```
sozercan: can we add this to retag using `registry.k8s.io/csi-secrets-store/driver` or ideally find an already retagged image that's patchable and has multi platform support

ashnamehrotra: looks like `registry.k8s.io/csi-secrets-store/driver:v1.4.8` exists, changed it to that. I'm not sure what you mean by add to retag?

pkg/patch/patch.go
```diff
@@ -887,3 +927,70 @@ func patchMultiPlatformImage(
...
+		return nil, fmt.Errorf("error parsing reference %q: %w", imageRef, err)
+	}
+
+	desc, err := remote.Get(ref)
+	if err != nil {
+		return nil, fmt.Errorf("error fetching descriptor for %q: %w", imageRef, err)
+	}
+
+	if !desc.MediaType.IsIndex() {
+		return nil, fmt.Errorf("expected multi-arch image but got single-arch image")
```
sozercan:
```suggestion
		return nil, fmt.Errorf("expected multi-platform image but got single platform image")
```

integration/multiarch/patch_test.go
```diff
@@ -126,6 +126,158 @@ func TestPatch(t *testing.T) {
...
+	require.True(t, platformSet["linux/arm64"], "should contain linux/arm64")
+
+	// Should have Windows platforms with different OS versions
+	windowsCount := 0
+	for key := range platformSet {
+		if strings.HasPrefix(key, "windows/amd64:") {
+			windowsCount++
+		}
+	}
+	require.Equal(t, 2, windowsCount, "should contain 2 windows/amd64 variants with different OS versions")
```
sozercan: can we get the number info from the registry instead of hardcoding it

integration/multiarch/patch_test.go
```diff
@@ -126,6 +126,158 @@ func TestPatch(t *testing.T) {
...
+	// Get patched manifest platforms
+	patchedPlatforms := getManifestPlatforms(t, patchedRef)
+	t.Logf("patched platforms: %v", patchedPlatforms)
+
+	// Verify all original platforms are preserved
+	require.Equal(t, len(originalPlatforms), len(patchedPlatforms),
+		"patched manifest should have same number of platforms as original")
+
+	// Check that we have the expected 4 platforms (2 Linux + 2 Windows)
+	require.Equal(t, 4, len(patchedPlatforms),
```
sozercan: same here

pkg/patch/patch.go
sozercan: nit: let's keep it consistent since it's not just an arch but diff platforms like windows

sozercan:
@ashnamehrotra can we remove the "unknown/unknown" digests in the image? https://oci.dag.dev/?image=nginx%3A1.27.0

image

sozercan (approved):
great job! lgtm

PR #1141, opened 2025-06-27
feat: support update all for multi platform patching

### LINKED ISSUE #1071: [REQ] Multi-Arch - Support update all features

None

If report-dir is not provided with multi arch, copa should patch the architectures that exist. Add a --platform flag to be able to differentiate.

ashnamehrotra:
- Checks if manifest list or single arch image when no report is provided
- if no reportDir provided and is multi platform image, by default patches all platforms unless `--platforms` flag is set to filter which ones to patch

Closes #1071


website/docs/multiplatform-patching.md
```diff
@@ -27,31 +35,70 @@ trivy image --vuln-type os --scanners vuln --ignore-unfixed \
...
+
+### Method 2: Update all outdated packages
 
-To patch a multi-platform image, you can use the `copa patch` command with the `--image` flag to specify the multi-platform image, the `--report` flag to point to the directory containing your vulnerability reports, and optionally a `--tag` for the final patched image.
+To patch a multi-platform across all platforms and update all outdated packages to latest, use the `copa patch` command with the `--image` flag to specify the multi-platform image, and omit the `--report` flag. 
 
 ```bash
 copa patch \
   --image $IMAGE \
   --report reports
```
robert-cronin: should the reports flag here be removed for the update all example?

website/docs/multiplatform-patching.md
```diff
@@ -27,31 +35,70 @@ trivy image --vuln-type os --scanners vuln --ignore-unfixed \
...
 ```
+To filter only specific platforms to patch, provide them with the `--platform` flag.
 
-Key Flags for Multi-Platform Patching:
+```bash
+copa patch \
+  --image $IMAGE \
+  --report reports
```
robert-cronin: same here for the update all example

website/docs/multiplatform-patching.md
```diff
@@ -27,31 +35,70 @@ trivy image --vuln-type os --scanners vuln --ignore-unfixed \
...
+   --platform=linux/amd64,linux/arm64
+```
 
-- `--report <directory_path>`: Specifies the directory containing platform-specific vulnerability reports.
+#### Valid Platform Values
+
+The `--platform` flag accepts the following values:
```
robert-cronin, line 66: I think this should be `--platforms` not `--platform`, looking at cmd.go

pkg/patch/patch.go
```diff
@@ -719,20 +760,106 @@ func getRepoNameWithDigest(patchedImageName, imageDigest string) string {
...
+			// Process all platforms, marking which should be patched vs preserved
+			for _, p := range discoveredPlatforms {
+				platformCopy := p
+				key := buildkit.PlatformKey(p.Platform)
+				if shouldPatchMap[key] {
+					// Platform should be patched with all
+					platformCopy.ReportFile = ""
+				} else {
+					// Platform should be preserved
+					platformCopy.ReportFile = "PRESERVE"
```
robert-cronin: nit: it might be better to add a field like `ShouldPreserve bool` to PatchPlatform

pkg/patch/patch.go
```diff
@@ -719,20 +760,106 @@ func getRepoNameWithDigest(patchedImageName, imageDigest string) string {
...
+	"linux/s390x", "linux/386", "linux/arm/v7", "linux/arm/v6",
+}
+
+// filterPlatforms filters discovered platforms based on user-specified target platforms.
+func filterPlatforms(discoveredPlatforms []types.PatchPlatform, targetPlatforms []string) []types.PatchPlatform {
+	var filtered []types.PatchPlatform
+
+	for _, target := range targetPlatforms {
+		// Validate platform against allowed list
+		if !slices.Contains(validPlatforms, target) {
```
robert-cronin, line 911: nit: `platforms.Parse` handles invalid platform strings so I think this is redundant

pkg/patch/patch.go
```diff
@@ -719,20 +760,106 @@ func getRepoNameWithDigest(patchedImageName, imageDigest string) string {
 	return nameWithDigest
 }
 
+var validPlatforms = []string{
```
robert-cronin, line 892: is this validation specific to copa (i.e. we only patch these platforms) or a guard against bad platform strings?

ashnamehrotra: this was to validate that we are only accepting linux based platforms that we are able to emulate. Is there any I am missing from this list?

pkg/patch/patch.go
ashnamehrotra: I can remove this if we want to remove validPlatforms (from the comment below)

integration/singlearch/patch_test.go
```diff
@@ -107,6 +107,14 @@ func TestPatch(t *testing.T) {
 			t.Log("patching image")
 			patch(t, ref, tagPatched, dir, img.IgnoreErrors, reportFile)
 
+			// For no-report tests with manifest images, Copa creates platform-specific tags like "-patched-amd64"
+			// The scanning should look for the tag that Copa actually created
+			scanTag := tagPatched
+			if !reportFile && img.IsManifest {
```
sozercan: instead of adding the `isManifest` flag to every test image, can we inspect the type using docker automatically?

integration/singlearch/patch_test.go
```diff
@@ -176,6 +184,11 @@ func patch(t *testing.T, ref, patchedTag, path string, ignoreErrors bool, report
 		reportPath = "-r=" + path + "/scan.json"
 	}
 
+	var platformsFlag string
+	if !reportFile {
+		platformsFlag = "--platforms=linux/amd64"
```
sozercan: are we doing this because tests take too long?

pkg/patch/cmd.go
```diff
@@ -76,6 +78,10 @@ func NewPatchCmd() *cobra.Command {
 	flags.StringVarP(&ua.format, "format", "f", "openvex", "Output format, defaults to 'openvex'")
 	flags.StringVarP(&ua.output, "output", "o", "", "Output file path")
 	flags.BoolVarP(&ua.push, "push", "p", false, "Push patched image to destination registry")
+	flags.StringSliceVar(&ua.platform, "platforms", nil,
+		"Target platform(s) for multi-arch images when no report directory is provided (e.g., linux/amd64,linux/arm64). "+
+			"Valid platforms: linux/amd64, linux/arm64, linux/riscv64, linux/ppc64le, linux/s390x, linux/386, linux/arm/v7, linux/arm/v6. "+
+			"If platform flag is used, only specified platforms are patched and the rest are preserved. If not specified, all platforms are patched.")
```
sozercan:
```suggestion
			"If platform flag is used, only specified platforms are patched and the rest are preserved. If not specified, all platforms present in the image are patched.")
```

pkg/patch/cmd.go
```diff
@@ -76,6 +78,10 @@ func NewPatchCmd() *cobra.Command {
 	flags.StringVarP(&ua.format, "format", "f", "openvex", "Output format, defaults to 'openvex'")
 	flags.StringVarP(&ua.output, "output", "o", "", "Output file path")
 	flags.BoolVarP(&ua.push, "push", "p", false, "Push patched image to destination registry")
+	flags.StringSliceVar(&ua.platform, "platforms", nil,
```
sozercan:
```suggestion
	flags.StringSliceVar(&ua.platform, "platform", nil,
```

to align with Docker CLI https://docs.docker.com/build/building/multi-platform/

pkg/patch/patch.go
sozercan: there is a list here but it's not exported: https://github.com/docker/cli/blob/74a896f18c9f8b10ce8a721f01c9f6fb71584987/cli/command/completion/functions.go#L154

sozercan: looks like we are missing `"linux/arm64/v8"`, and we'll need to preserve the wasm one

website/docs/multiplatform-patching.md
```diff
@@ -27,31 +35,68 @@ trivy image --vuln-type os --scanners vuln --ignore-unfixed \
...
 - `--tag <final_tag>` (optional): The tag for the final, reassembled multi-platform manifest (e.g., `1.0-patched`).
 - `--push` (optional): If included, Copa pushes the final multi-platform manifest to the registry.
 - `--ignore-errors` (optional, default: `false`): When `false` (default), Copa warns about errors and fails if any platform encounters an error. When `true`, Copa warns about errors but continues processing other platforms.
 
 ### Things to Keep in Mind
 
-If you don't provide a `--report` flag pointing to a directory, Copa will not perform multi-platform patching and will instead only patch the image for the architecture of the host machine.
-
-If `--push` is not specified, the individual patched images will be saved locally, and you can push them to your registry later using `docker push` and then `docker manifest create/push` to create the multi-platform manifest.
+- **Automatic platform detection**: Copa automatically detects whether an image is multi-platform (manifest list) or single-platform and handles them accordingly.
```
sozercan, line 89:
```suggestion
- **Automatic platform detection**: Copa automatically detects whether an image is multi-platform (Docker manifest list or OCI index) or single-platform and handles them accordingly.
```

pkg/patch/patch.go
```diff
@@ -222,7 +264,10 @@ func patchWithContext(
 	if f.IsDir() {
 		// Handle directory - multi-platform patching
 		log.Debugf("Using report directory: %s", reportPath)
-		return patchMultiPlatformImage(ctx, ch, image, reportPath, patchedTag, suffix, workingFolder, scanner, format, output, loader, ignoreError, push, bkOpts)
+		if len(targetPlatforms) > 0 {
+			log.Warnf("Platform flag ignored when report directory is provided")
```
sozercan: should we categorize this as informational, not a warning?

pkg/patch/cmd.go
sozercan: please update docs with this change too

integration/singlearch/patch_test.go
ashnamehrotra: we can, but we would still end up hardcoding a value since we won't be able to do `docker manifest inspect` for the local images in the test

integration/singlearch/patch_test.go
ashnamehrotra: I added this because this workflow is a single architecture test so I was thinking we can patch/scan for the workflow platform? it would also increase test time if we wanted to patch all platforms

pkg/patch/patch.go
ashnamehrotra: updated to match this list - 	"linux/riscv64", "windows/amd64", and "wasip1/wasm" will be preserved since they will be filtered out of validPlatforms

robert-cronin (approved):
LGTM 👍

integration/singlearch/fixtures/test-images.json
```diff
@@ -14,15 +15,17 @@
...
     },
     {
         "image": "docker.io/library/nginx",
         "tag": "1.21.6",
         "digest": "sha256:2bcabc23b45489fb0885d69a06ba1d648aeda973fae7bb981bafbb884165e514",
         "distro": "Debian",
         "description": "Valid dpkg/status, apt present",
-        "ignoreErrors": false
+        "ignoreErrors": false,
+        "isManifest": true
```
sozercan: this is a manifest list, why do we mark as manifest?
https://oci.dag.dev/?image=docker.io%2Flibrary%2Fnginx%40sha256%3A2bcabc23b45489fb0885d69a06ba1d648aeda973fae7bb981bafbb884165e514

sozercan: do we mean `isManifestList` here?

ashnamehrotra: yes, should be isManifestList. Updated this!

PR #1197, opened 2025-07-09
fix: Get host platform info in single arch integration test

### LINKED ISSUE #1196: [BUG] Check for platforms in single arch integration test
### Version of copa

_No response_

### Expected Behavior

Test is assuming amd64 architecture exists in integration test:
https://github.com/project-copacetic/copacetic/blob/main/integration/singlearch/patch_test.go#L110-L114

Better approach would be to check for a platform instead of hardcoding this value.

### Actual Behavior

_No response_

### Steps To Reproduce

_No response_

amanycodes:
used containerd to get host platform info, was thinking of using `docker info` but then cross build might have caused issue that and made the test more brittle, were using containerd for getting platform info in patching flow aswell.

Closes #1196


integration/singlearch/patch_test.go
```diff
@@ -111,7 +112,10 @@ func TestPatch(t *testing.T) {
 			// The scanning should look for the tag that Copa actually created
 			scanTag := tagPatched
 			if !reportFile && img.IsManifestList {
-				scanTag += "-amd64"
+				// Determine the host platform
+				hostPlatform := platforms.DefaultSpec()
+				platformArch := fmt.Sprintf("-%s", hostPlatform.Architecture)
+				scanTag += platformArch
```
ashnamehrotra: Instead of host platform which would always be `amd64` in our CI, I was thinking we could check for one of the present platforms in the image we are patching. For example, if we are patching `docker.io/arm64v8/redis:7.4.0-alpine`, there is no `amd64` platform in the manifest list, so we should check for the `arm64` platform tag that exists rather than `amd64`

amanycodes:
there was some issues with the commits, i  pushed the changes and they were not visible here, opening it again. thanks!

integration/singlearch/patch_test.go
amanycodes: we are now getting the platform info from the manifest list and if it is the same as the hostplatform we take it or else we raise an error(we can just keep the manifest list platform if that's needed but there might be multiple in case of a different example than `docker.io/arm64v8/redis:7.4.0-alpine`). please let me know if this is what you needed!

amanycodes:
@ashnamehrotra hmm the tests are failing, so for images which have mulitiple architectures in the manifest list, what should we do about them in this new logic we're thinking? apparently some images don't have a manifest list and we were tagging them amd64 earlier which was correct? 
in a non-host image case like for `docker.io/arm64v8/redis:7.4.0-alpine` running on arm64 but out host machine running on amd64 we want to tag it -arm64 but this is a really specific case here right? the --platform tag is for multiarch images?
either we add platform to each test image in the json and then update the checking logic.
WDYT!

@sozercan would love to know your thoughts on this aswell, the new non-host patch logic would help me understand the case better!

ashnamehrotra:
@amanycodes  If the image is not a manifest list it is just tagged with "-patched", but if it is a manifest list the manifest is tagged "-patched" and the individual image platform references are tagged with "-patched-arch". So we want to check for one of the expected platform tags to make sure it exists and has patched vulnerabilities. For `docker.io/arm64v8/redis:7.4.0-alpine`, we would check for "-patched-arm64" existing instead of "-patched-amd64". Let me know if that helps explain it!

amanycodes:
@ashnamehrotra thanks for the explanation! just one last doubt, what happens when the manifest image has multiple architectures? and none of them is the same as our host platform? `eg: registry.k8s.io/kube-proxy (Custom Google Distroless)` since this was failing. which one do we choose and what is tagged then?

ashnamehrotra:
@amanycodes since it is multi-platform patching, it will patch the architectures that exist in the manifest using qemu, so it doesn't need to match the host platform. Since it is for testing, we can choose one of the architectures to check for. Before we were just checking amd64 by default, but we want to change that and use a platform that we know is in the list since there is a chance amd64 is not present.

amanycodes:
@ashnamehrotra okay, great! the issue is very clear for me now. thanks alot for the clarification!

ashnamehrotra:
@amanycodes it looks like for fluent bit the tag is erroring with `docker.io/fluent/fluent-bit:1.8.4-patched-arm` instead of `docker.io/fluent/fluent-bit:1.8.4-patched-arm64`

amanycodes:
@ashnamehrotra I updated the logic so that now we will filter the platforms which are supported by copa and then patch the first from that list if multiple, but the fluent image is erroring in the patching process?
https://github.com/project-copacetic/copacetic/actions/runs/16269222759/job/45932351568?pr=1197#step:10:387
is this because of the wrong tag? or some flakiness?

ashnamehrotra:
@amanycodes looks like Debian buster reached EOL last year, the grace period may have ended and it doesn't seem specific to your PR. Could we try removing the image from the test?

amanycodes:
@ashnamehrotra seems like you raised a pr fixing the eol image. will wait for that to merge. thanks for the help!

amanycodes:
@ashnamehrotra this is good to go now?

integration/singlearch/patch_test.go
```diff
@@ -145,6 +159,46 @@ func TestPatch(t *testing.T) {
 	}
 }
 
+func getManifestPlatforms(t *testing.T, imageRef string) []manifestPlatform {
+	validPlatforms := map[string]bool{
```
ashnamehrotra, line 171: can we have this value match the value of `validPlatforms` in patch.go?

integration/singlearch/patch_test.go
```diff
@@ -111,7 +118,14 @@ func TestPatch(t *testing.T) {
 			// The scanning should look for the tag that Copa actually created
 			scanTag := tagPatched
 			if !reportFile && img.IsManifestList {
-				scanTag += "-amd64"
+				hostPlatform := platforms.DefaultSpec().Architecture
+				imagePlatforms := getManifestPlatforms(t, ref)
+
+				targetArch := hostPlatform
+				if len(imagePlatforms) > 0 && imagePlatforms[0].Architecture != hostPlatform {
```
ashnamehrotra: this works, but for clarity maybe we can change this to:

if host platform isn't in the list of validPlatforms, then use `imagePlatforms[0].Architecture` as target arch?
in most cases hostPlatform (linux/amd64) will be present, and theres a chance it is not at imagePlatforms[0]

ashnamehrotra (changes requested):

ashnamehrotra (approved):
LGTM, thanks!

PR #1219, opened 2025-07-21
chore: Exempt good first issues from getting stale

amanycodes:
Added `good first issue` in the exempted issue labels for the stale bot.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -31,6 +31,6 @@ jobs:
             This pull request has been automatically marked as stale because it has not had 
             recent activity. It will be closed if no further activity occurs. 
             Thank you for your contributions.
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yml
```diff
@@ -31,6 +31,6 @@ jobs:
             This pull request has been automatically marked as stale because it has not had 
             recent activity. It will be closed if no further activity occurs. 
             Thank you for your contributions.
-          exempt-issue-labels: 'security,bug'
+          exempt-issue-labels: 'security,bug,good first issue'
```
sozercan: since users can assign labels, can we remove bug from this list? 

let's add a new entry called `triaged` so maintainers can specify a label to keep issues open

PR #1220, opened 2025-07-21
fix: add almalinux to pkgmanager os type check

ashnamehrotra:
Adds check for "almalinux" in addition to "alma" as a supported osType check for package manager.

Closes #<_issue_ID_>


pkg/pkgmgr/pkgmgr.go
```diff
@@ -44,7 +44,7 @@ func GetPackageManager(osType string, osVersion string, config *buildkit.Config,
 			osVersion:     osVersion,
 			osType:        osType,
 		}, nil
-	case "cbl-mariner", "azurelinux", "centos", "oracle", "redhat", "rocky", "amazon", "alma":
+	case "cbl-mariner", "azurelinux", "centos", "oracle", "redhat", "rocky", "amazon", "alma", "almalinux":
```
sozercan, line 47: does this need to be in https://github.com/project-copacetic/copacetic/blob/25e58fcd843e2b701b82eda77e098da99cfa2b92/pkg/patch/patch.go#L867-L868?

ashnamehrotra: this function only called when scan report isn't passed in, otherwise we get the osType from the scan report. 
Since this was qualys specific I didn't add the check to `getOSType`.

ashnamehrotra: I think if the osType is "almalinux" in the release file this would still work since we use contains

PR #1253, opened 2025-08-12
fix: make GetUniqueLatestUpdates deterministic

acortelyou:
fix: make GetUniqueLatestUpdates deterministic

We observed that running copa against the same image multiple times was producing output with varying hashes.

Investigation revealed that the order of the "tdnf -y install <pkgs>" command was non-deterministic and changed from run to run.

This change sorts the results of the package manager utility function `GetUniqueLatestUpdates` so that generated actions will stay consistent.

The tests for `pkgmgr` have been updated to confirm this behavior.

### DIFF
--- a/integration/multiarch/fixtures/test-images.json
+++ b/integration/multiarch/fixtures/test-images.json
@@ -9,9 +9,7 @@
     "push": true,
     "platforms": [
       "linux/amd64",
-      "linux/arm/v5",
       "linux/386",
-      "linux/mips64le",
       "linux/s390x",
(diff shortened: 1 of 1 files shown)


acortelyou:
The failing checks don't seem related to these changes: I see other PRs are failing in the same way at the moment.

ashnamehrotra (approved):
LGTM

robert-cronin (approved):
I am looking into the failing test now, I think this is okay to merge once that has been resolved 👍

PR #1274, opened 2025-08-27
fix: Non Zero Exit code for no upgradable packages causes build fail

### LINKED ISSUE #1223: [REQ] Fix Error Output for no packages to update

None

Would like Copacetic to return a specific error code + message when no there's no packages to update. This is to clear up confusion around the current error code returned despite there being no errors found. 

Link to issue #802 

amanycodes:
The package manager shell commands now always exit with code 0. A marker file is now created within the build context only if upgradable packages are detected. That is read and if no upgradable packages found the sentinel error propagates to `patchSingleArchImage` and it catches this error and logs the user that the package is not upgradable (in summary table if multi-arch setting).

Attempt to close #1223


amanycodes:
@ashnamehrotra @robert-cronin I think the pipeline looks fine. The sentinel error is being outputted but the build is still failing when no upgradable packages are present, even though now the shell give exit code 0.  I'm confused as to whether I'm not catching the sentinel error at the correct point in the control flow?  Would love to have some eyes on this, Thanks!

ashnamehrotra:
@amanycodes can we add a CI test for this? 

We still want the error on the copa side which is correct, since we do not want to create a "patched" tag when the image is unchanged. We just want to have the exit code changed to "0" so that when copa is running in a pipeline it wouldnt cause a CI failure. Does that make sense?

ashnamehrotra:
@amanycodes are you still working on this?

pkg/common/errors.go
```diff
@@ -0,0 +1,5 @@
+package common
+
+import "errors"
+
+var ErrNoUpdatesAvailable = errors.New("no upgradable packages")
```
robert-cronin: Is this being used anywhere? If not, I think we can remove it

amanycodes:
Hi @ashnamehrotra , I'm back from midsems, and have started working on it.

pkg/common/errors.go
amanycodes: this was being used as the sentinel error but due to cyclic deps i moved error.go to /types. i hope that is okay.

sozercan:
@amanycodes looks like there are some test failures, can you ptal?

PR #1275, opened 2025-08-28
fix: get platform for rpm tooling image

### LINKED ISSUE #1262: [BUG] Copa downloads x86_64 arch packages for arm64 image
### Version of copa

_No response_

### Expected Behavior

When patching an arm64 image on amd64 architecture (ex `sozercan/test:test2`), copa logs indicate x86_64 packages downloaded

### Actual Behavior

_No response_

### Steps To Reproduce

_No response_

ashnamehrotra:
Set platform RPM tooling image to the platform of image to be patched. Currently it defaults to host platform.

Closes #1262

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -20,7 +20,7 @@ require (
 	github.com/moby/buildkit v0.24.0
 	github.com/opencontainers/go-digest v1.0.0
 	github.com/opencontainers/image-spec v1.1.1
-	github.com/openvex/go-vex v0.2.5
+	github.com/openvex/go-vex v0.2.6
 	github.com/quay/claircore v1.5.39
 	github.com/sirupsen/logrus v1.9.3
 	github.com/spf13/cobra v1.9.1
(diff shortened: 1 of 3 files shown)


ashnamehrotra:
@robert-cronin @sozercan do we want to add a fallback tooling image with host platform in this pr? it could complicated the CLI a bit to add a flag for this. So far we haven't had issues with the tooling images not existing for arm64/amd64. If we do this, we should add the same for [debian pkgmgr](https://github.com/project-copacetic/copacetic/blob/main/pkg/pkgmgr/dpkg.go#L427-L437).

robert-cronin (approved):

sozercan:
@ashnamehrotra are there any tooling images that doesn't support all the platforms copa supports? (not just amd64 and arm64)

ashnamehrotra:
@sozercan 
Yes that is a good point! looking at our tooling images some platforms like `riscv64` are missing. https://github.com/project-copacetic/copacetic/pkgs/container/copacetic%2Fdebian. I will add a fallback to host platform in this case. I'm not sure if we need a flag for this configuration on the user side though? Or we can have the error ignored via `--ignore-errors` in which case it falls back to host platform?

sozercan:
@ashnamehrotra i am fine not adding a user facing flag, and just falling back to host platform in case target platform image is missing/not exists

sozercan (approved):

PR #1280, opened 2025-09-03
fix: add fallback for baseImage not found

### LINKED ISSUE #1107: [BUG] Copa fails when the image referenced in config label `BaseImage` is not available
### Version of copa

Release 0.9.0

### Expected Behavior

When Copa patches an image it sets a label `BaseImage` in Config -> Labels of the newly patched image. If this referenced image is removed Copa will fail on the next time it tries to patch the image.

For example, in the scenario of patching `my-reg.example.com/my-repo/my-image:v1.0.0`, with a patched tag of `v1.0.0-patched-111`, the resulting image would contain a BaseImage label with value `my-reg.example.com/my-repo/my-image:v1.0.0` (assumes some vulnerabilities were patched).

We then run Copa again on `my-reg.example.com/my-repo/my-image:v1.0.0-patched-111` using a new patch tag of `v1.0.0-patched-222` and again it patches the image due to vulnerabilities. The newly created image would still have a BaseImage label of `my-reg.example.com/my-repo/my-image:v1.0.0`. We continue this loop and it remains the same. Note, we are always trying to patch an already patched image as that patched image may now have vulnerabilities. If we always patch the initial image we will always fix the same vulnerabilities again and again.

If at some point we remove the image `my-reg.example.com/my-repo/my-image:v1.0.0` then subsequent patching attempts fail.

Why is Copa setting the BaseImage label and why does it set it to the value of the initial image patched? Why is it not updating the BaseImage label to the value of the image that we patched?

### Actual Behavior

_No response_

### Steps To Reproduce

_No response_

ashnamehrotra:
If BaseImage is not found, Copa will still proceed with patching.

1. `ashnam/nginx:1.21.6-arm64-patched` was created by paching `ashnam/nginx:1.21.6-arm64` via Trivy report. 
2. After removing baseImage `ashnam/nginx:1.21.6-arm64` from the registry, Copa would not be able to patch `ashnam/nginx:1.21.6-arm64-patched` again with update all.
3. After these changes, Copa continues to patch `ashnam/nginx:1.21.6-arm64-patched` 

Closes #1107


robert-cronin (approved):
I think apart from the issues copilot raised, this looks okay to me

PR #1399, opened 2025-11-26
feat: add test environment utilities for BuildKit integration

### LINKED ISSUE #1054: Integration testing
There is a simple e2e test that executes the copa CLI.

I think we'd benefit greatly from more integration level tests that can just execute go code and inspect outputs.
Probably this means some refactoring where instead of taking a plain `buildkit.Client` we need to take a gateway client.

We basically do a *lot* of this in https://github.com/Azure/dalec.
https://pkg.go.dev/github.com/Azure/dalec/test/testenv?utm_source=godoc#BuildxEnv.RunTest

By utilizing the gateway client we can write tests that just look at the state of the content before its been exported to the client (just like as if you were code running inside copa itself).
One thing it cannot do is look at individual layers, so it'd still need to export that content out to have tests to check things like the number of layers are what is expected.

robert-cronin:
Closes #1054

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -37,7 +37,7 @@ jobs:
     timeout-minutes: 5
     steps:
       - name: Harden Runner
-        uses: step-security/harden-runner@e3f713f2d8f53843e71c69a996d56f51aa9adfb9 # v2.3.1
+        uses: step-security/harden-runner@5ef0c079ce82195b2a36a210272d6b661572d83e # v2.3.1
         with:
           egress-policy: audit
       - name: Check out code
@@ -68,7 +68,7 @@ jobs:
       contents: read
     steps:
       - name: Harden Runner
-        uses: step-security/harden-runner@e3f713f2d8f53843e71c69a996d56f51aa9adfb9 # v2.3.1
+        uses: step-security/harden-runner@5ef0c079ce82195b2a36a210272d6b661572d83e # v2.3.1
         with:
           egress-policy: audit
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
@@ -275,7 +275,7 @@ jobs:
     timeout-minutes: 5
     steps:
       - name: Harden Runner
-        uses: step-security/harden-runner@e3f713f2d8f53843e71c69a996d56f51aa9adfb9 # v2.3.1
+        uses: step-security/harden-runner@5ef0c079ce82195b2a36a210272d6b661572d83e # v2.3.1
         with:
           egress-policy: audit
       - name: Check out code
--- a/.github/workflows/check-deps.yml
+++ b/.github/workflows/check-deps.yml
@@ -18,7 +18,7 @@ jobs:
     timeout-minutes: 10
     steps:
       - name: Harden the runner (Audit all outbound calls)
-        uses: step-security/harden-runner@e3f713f2d8f53843e71c69a996d56f51aa9adfb9 # v2.14.1
+        uses: step-security/harden-runner@5ef0c079ce82195b2a36a210272d6b661572d83e # v2.14.2
         with:
           egress-policy: audit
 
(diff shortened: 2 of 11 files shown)


test/testenv/layers.go
```diff
@@ -0,0 +1,339 @@
...
+		} `json:"layers"`
+	}
+
+	if err := json.Unmarshal(manifestData, &manifest); err != nil {
+		return nil, fmt.Errorf("failed to parse manifest: %w", err)
+	}
+
+	// Read config blob to get diff_ids
+	configDigest := manifest.Config.Digest
+	configBlobPath := filepath.Join(extractDir, "blobs", "sha256", configDigest[7:])
```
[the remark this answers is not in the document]
robert-cronin: Added parseDigest(), digestToPath() helpers

test/testenv/testenv.go
```diff
@@ -0,0 +1,169 @@
...
+	_, err := e.client.Build(ctx, solveOpt, "copa-test", func(ctx context.Context, c gwclient.Client) (*gwclient.Result, error) {
+		// Capture any test failures via panic recovery
+		defer func() {
+			if r := recover(); r != nil {
+				if err, ok := r.(error); ok {
+					testErr = err
+				} else {
+					testErr = fmt.Errorf("test panic: %v", r)
+				}
+			}
```
cpuguy83, line 117: Nice was just discussing a similar change in dalec.
Have you validated that the recover works?

integration/gateway/main_test.go
```diff
@@ -0,0 +1,44 @@
...
+	// testEnv is the shared test environment for all tests in this package.
+	testEnv *testenv.TestEnv
+
+	// buildkitAddr is the BuildKit address to use for tests.
+	// Can be set via -addr flag or COPA_BUILDKIT_ADDR environment variable.
+	buildkitAddr string
+)
+
+func TestMain(m *testing.M) {
+	flag.StringVar(&buildkitAddr, "addr", "", "BuildKit address (e.g., docker://, tcp://localhost:1234)")
```
cpuguy83, line 30: Should this use the same connection logic as the copoa bin?

test/testenv/testenv.go
robert-cronin: Thanks! Yep, pushed e953945 with unit tests. The recover catches panics from test code and converts them to clean test failures. It's defensive, not strictly required for t.Fatal, but useful if something unexpected panics inside the callback.

integration/gateway/main_test.go
robert-cronin: It uses the same buildkit.NewClient() function from pkg/buildkit/drivers.go. When addr is empty, it triggers autoClient() which tries docker, buildx, then default socket.

The tests now import the same dockercontainer and podmancontainer connection helpers as the binary (fixed in 22e036b). They don't register kubepod/nerdctl/ssh but those aren't used in CI.

test/testenv/layers.go
```diff
@@ -0,0 +1,392 @@
...
+	// Create temp directory for extraction
+	extractDir := filepath.Join(l.outputDir, "extract")
+	if err := os.MkdirAll(extractDir, 0o755); err != nil {
+		return nil, fmt.Errorf("failed to create extract directory: %w", err)
+	}
+	defer os.RemoveAll(extractDir)
+
+	// Extract tar using tar command
+	// This is simpler than pulling in archive/tar
+	if err := extractTar(tarPath, extractDir); err != nil {
```
cpuguy83: It looks like this is overkill for what's needed.
You could either just parse the tar file to get the index and manifests directly or use something like https://github.com/cpuguy83/tar2go which creates an `io/fs.FS` from an io.Reader.

No need to copy any extra data to disk.

robert-cronin: good point, switched the tar parsing to use an in memory fs.FS over the tar stream instead of extracting to disk and added t.Parallel() across the gateway tests

cpuguy83 (approved):

integration/gateway/distro_test.go
ashnamehrotra, line 1: nit should we rename this since we are testing debian and alpine too? LGTM otherwise!

robert-cronin: good call, updated this to better reflect whats in the test

PR #1420, opened 2025-12-18
fix: add registry authentication for platform discovery

sozercan:
Use authn.DefaultKeychain for remote registry calls in platform discovery
and preserved platform export functions. This fixes UNAUTHORIZED errors
when working with private registries like ACR, GCR, ECR, etc.

The fix adds authentication to:
- DiscoverPlatformsFromReference()
- exportPreservedPlatformsToOutput()
- exportOriginalImagePlatformsAsOCI()

Credentials are automatically read from Docker config (~/.docker/config.json)
and credential helpers.

Closes #<_issue_ID_>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -22,7 +22,7 @@ require (
 	github.com/knqyf263/go-apk-version v0.0.0-20200609155635-041fdbb8563f
 	github.com/knqyf263/go-deb-version v0.0.0-20241115132648-6f4aee6ccd23
 	github.com/knqyf263/go-rpm-version v0.0.0-20220614171824-631e686d1075
-	github.com/moby/buildkit v0.26.2
+	github.com/moby/buildkit v0.26.3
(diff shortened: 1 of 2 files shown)


.github/workflows/private-registry-test.yml
```diff
@@ -0,0 +1,106 @@
+name: "[Blocking] Private Registry E2E Test"
+
+# This workflow tests Copa's ability to work with private container registries.
+# It runs only on merge to main (not on PRs) because it requires access to
```
robert-cronin, line 4: I think the current trigger will run on pushes to any branch, perhaps we can restrict it to main only?
```
  on:                                                                                                    
    push:                                                                                                
      branches:                                                                                          
        - main                                                                                           
      paths-ignore: ...
```

sozercan: thanks, i removed for testing, added back

PR #1453, opened 2026-01-28
feat: add demo recordings and asciinema player to website

### LINKED ISSUE #1452: [BUG] .NET patching hangs without --debug flag
### Version of copa

v0.0.13

### Expected Behavior

`copa patch` completes .NET image patching in ~18 seconds (as it does with `--debug`).

### Actual Behavior

`copa patch` hangs indefinitely after the DLL patching phase completes, eventually timing out. Adding `--debug` makes the same operation succeed in ~18 seconds.

The root cause is the BuildKit progress channel is unbuffered. .NET patching generates heavy output (dotnet restore, DLL copying, deps.json updates) which causes backpressure against the progrock TUI. `--debug` uses PlainMode which writes directly to stderr and never blocks.

### Steps To Reproduce

_No response_

robert-cronin:
- Add demo-magic scripts for .NET, Node.js (Ghost CMS), and Python app-level patching demos
- Add AsciinemaPlayer React component and demos page to the Docusaurus website with embedded cast recordings
- Add cleanup scripts for each demo
- Fix unbuffered BuildKit progress channel causing .NET patching to hang without `--debug`

Closes #1452

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -39,7 +39,7 @@ require (
 	golang.org/x/exp v0.0.0-20250911091902-df9299821621
 	golang.org/x/sync v0.19.0
 	golang.org/x/term v0.39.0
-	google.golang.org/grpc v1.78.0
+	google.golang.org/grpc v1.79.1
 	gopkg.in/yaml.v3 v3.0.1
 	k8s.io/apimachinery v0.35.1
 )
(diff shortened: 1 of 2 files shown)


ashnamehrotra:
@robert-cronin I wasnt able to find a published and vulnerable dot net image, but if we build this one locally it has vulnerabilities that can be patched https://github.com/Azure/azure-relay-bridge/blob/693a39570f095750bcc1900a744c18959b78c5eb/Dockerfile. Maybe we can replace my image with building and patching this in the demo?

robert-cronin:
Thanks @ashnamehrotra! Updated the .NET demo to build from Azure Relay Bridge locally as you suggested. Used docker:// to avoid BuildKit container driver issues with local-only images. Also switched Python to python:3.11-alpine and removed the Node.js demo due to #1462.

website/static/casts/.gitkeep
ashnamehrotra, line 1: do we need this file?

ashnamehrotra: and should we add node js cast here as well?

demo/README.md
```diff
@@ -1,3 +1,50 @@
...
+| `copa-demo-nodejs.sh` | `copa-demo-nodejs-cleanup.sh` |
+| `copa-demo-python.sh` | `copa-demo-python-cleanup.sh` |
+
+```bash
+bash copa-demo-python-cleanup.sh
+```
+
+## Recording
+
+To record a demo as an asciinema cast for the website:
```
ashnamehrotra: I think we can get rid of this section?

website/static/casts/.gitkeep
robert-cronin: Dropped this gitkeep file, Also, I've removed the Node.js demo for now because npm override patching was increasing vulnerability counts instead of reducing them. Tracked in #1462. Happy to redo it once that issue is resolved.

demo/README.md
robert-cronin: done

PR #1468, opened 2026-02-13
feat: add patch summary output for vulnerabilities

### LINKED ISSUE #1451: [REQ] Add patch summary output showing total/patched/skipped vulnerabilities

## Summary

Copa currently outputs only the patched image reference after a successful patch:

Patched image (linux/arm64): docker.io/example/image:patched

There is no summary of what was patched, what was skipped, or why. The report parser at `pkg/report/trivy.go` silently filters out vulnerabilities with no available fix (`FixedVersion == ""`), giving users no visibility into the patching results.

## Proposed Output

After patching, Copa should print a summary like:

```
Patch Summary:
  Total vulnerabilities in report: 24
  Patched:  6 (5 OS, 1 library)
  Skipped: 18 (no fix available)

Patched image (linux/arm64): docker.io/example/image:patched
```

## Why

Without this, users must re-scan with Trivy to understand what was and wasn't patched and may misinterpret remaining vulnerabilities as a Copa failure rather than upstream packages with no fix available.

cnaples79:
## Summary
- Add patch summary output showing total vulnerabilities in report, patched count, and skipped-no-fix count.
- Include OS/library patched breakdown in summary output.
- Add Trivy summary logic and focused tests for counting behavior.

## Rationale
- Issue #1451 asks for visibility into what Copa patched versus skipped due to missing fixes, instead of only printing patched image references.

## Changes
- `pkg/report/report.go`
  - Added `PatchSummary` and `TrySummarizeScanReport`.
- `pkg/report/trivy.go`
  - Added summary extraction logic for Trivy reports.
- `pkg/patch/single.go`
  - Added summary logging in report-based patch flows.
- `pkg/report/trivy_test.go`
  - Added tests for summary counts and pkg-type filtering.

## Validation
- CI passing on fork

Fixes #1451


pkg/report/report.go
```diff
@@ -29,13 +29,31 @@ type ScanReportParser interface {
 	ParseWithLibraryPatchLevel(string, string) (*unversioned.UpdateManifest, error)
 }
 
+// PatchSummary captures vulnerability patching visibility for a scan report.
+type PatchSummary struct {
+	TotalVulnerabilities int
```
ashnamehrotra, line 34: Can we make this field optional for the summary? In the case the user choses to update without a report it should still show a patched packages count.

ashnamehrotra (changes requested):

pkg/patch/single.go
```diff
@@ -116,7 +116,13 @@ func patchSingleArchImage(
 
 	// Parse report for update packages
 	var updates *unversioned.UpdateManifest
+	var patchSummary *report.PatchSummary
 	if reportFile != "" {
+		patchSummary, err = report.TrySummarizeScanReport(reportFile, scanner, pkgTypes)
+		if err != nil {
+			return nil, err
```
robert-cronin, line 124: This error path aborts the entire patch if summary generation fails. Since the summary is purely informational, can we log the error as a warning and continue with the patch instead? Something like:

patchSummary, err = report.TrySummarizeScanReport(reportFile, scanner, pkgTypes)
if err != nil {
    log.Warnf("Failed to generate patch summary: %v", err)
}

Otherwise a summary-only regression (e.g., future schema change) would prevent images from being patched.

robert-cronin (commented):
Great work, thanks for the PR! I tested locally and the feature works for me. One thing i noticed is that the summary error at single.go:121 is fatal, see inline comment. A couple of minor notes for follow up: 
1. "Patched" counts CVEs not unique packages, which may confuse users, and 
2. for multi-platform images the summary prints once per arch with no platform label.

omercnet:
Thanks for opening this — you identified the right problem and got the conversation started. #1517 has since incorporated all the review feedback from this PR (double-parse, error handling, patch-level correctness, library overcounting) and is already approved. I think this can be closed in favor of #1517. Your `PatchSummary` concept is essentially the same struct that #1517 uses, so the core design carries over.

ashnamehrotra:
completed in #1517

PR #1475, opened 2026-02-17
feat: add skip detection and cross-registry support for bulk patching

omercnet:
## Summary

This PR adds intelligent skip detection for bulk patching to avoid unnecessary re-patching when images have no fixable vulnerabilities, plus cross-registry workflow support for patching images from one registry and pushing to another.

### 1. Intelligent Skip Detection

Skip re-patching images when vulnerability reports show no fixable vulnerabilities, saving time and compute in scheduled/CI environments.

**Key Capabilities:**
- ArtifactName-based report matching (no filename convention required)
- Automatic version-bumped tags when re-patching is needed
- Scanner-agnostic: works with Trivy, native format, or custom plugins
- Fail-open behavior: missing/invalid reports → proceed with patching

**New Flags:**
```bash
--patched-reports-dir <dir>  # Directory with vulnerability reports
--force                      # Force re-patching even if no vulnerabilities
```

**Tag Versioning:**
When re-patching is needed, version-suffixed tags are used:
- `1.25.3-patched` ← initial patch
- `1.25.3-patched-1` ← first re-patch
- `1.25.3-patched-2` ← second re-patch

**Workflow Example:**
```bash
# 1. Patch images
copa patch --config bulk.yaml --push

# 2. Scan patched images (any filename works)
trivy image registry.io/nginx:1.25.3-patched -f json -o reports/nginx.json

# 3. Run with skip detection
copa patch --config bulk.yaml --push --patched-reports-dir ./reports
```

### 2. Cross-Registry Workflow Support

Patch images from one registry and push to another with proper skip detection and report matching.

**New Configuration:**
```yaml
apiVersion: copa.sh/v1alpha1
kind: PatchConfig

# Global default for all images
target:
  registry: "ghcr.io/myorg"

images:
  - name: "redis"
    image: "quay.io/opstree/redis"  # Source registry
    tags:
      strategy: "list"
      list: ["v8.2.1"]
    # Per-image override (optional)
    target:
      registry: "quay.io/special/redis"
```

**How It Works:**
- Copa extracts image name from source (`redis` from `quay.io/opstree/redis`)
- Appends to target registry → `ghcr.io/myorg/redis`
- Skip detection queries the correct target registry
- Reports with target registry ArtifactName match correctly
- Uses Docker credentials for authenticated registries

### Skip Detection
- **pkg/types/options.go**: Add `PatchedReportsDir` and `Force` fields
- **pkg/cmd/cmd.go**: Add `--patched-reports-dir` and `--force` flags
- **pkg/bulk/skip.go**: Core skip detection logic
  - `reportIndex` type for normalized ref → file path mapping
  - `buildReportIndex()` to scan directory and extract ArtifactName
  - `discoverExistingPatchTags()` to find patched versions
  - `evaluatePatchAction()` for skip decision workflow
  - Tag versioning utilities
- **pkg/bulk/engine.go**: Integrate skip detection into worker pipeline
- **pkg/bulk/skip_test.go**: Comprehensive unit tests
- **integration/bulk/**: Integration tests for skip workflow

### Cross-Registry Support
- **pkg/bulk/config.go**: Add `Target` field to `PatchConfig` and `Registry` to `TargetSpec`
- **pkg/bulk/engine.go**: Add `buildTargetRepository()` and `mergeTarget()` functions
- **pkg/bulk/discover.go**: Add Docker auth (`authn.DefaultKeychain`) to `listAllTags`
- **pkg/bulk/engine_test.go**: Tests for target repository building and merging

### Documentation
- **website/docs/bulk-image-patching.md**: Complete guide for both features
- **website/docs/troubleshooting.md**: Troubleshooting for skip detection and cross-registry
- **website/docs/faq.md**: Skip detection FAQ and tag versioning
- **website/docs/best-practices.md**: Best practices for scheduled patching

## Testing

All existing tests pass. New tests cover:
- Report indexing and ArtifactName extraction
- Tag versioning and discovery
- Skip decision workflow
- Target repository building with various image formats
- Global and per-image target merging
- Docker auth integration

## Benefits

- **Time Savings**: Skip unnecessary re-patching in scheduled workflows
- **User-Friendly**: No filename conventions required
- **Robust**: Handles reference normalization automatically
- **Flexible**: Works with any registry configuration
- **Secure**: Uses Docker credentials for authenticated registries
- **Scanner-Agnostic**: Works with Trivy, native format, or custom plugins
- **Fail-Safe**: Fail-open behavior ensures patches happen when in doubt

### Scheduled Patching
```bash
# Daily cron job
0 0 * * * copa patch --config bulk.yaml --push --patched-reports-dir ./reports
```

### Cross-Registry Mirroring
```yaml
target:
  registry: "ghcr.io/myorg"
images:
  - image: "quay.io/upstream/app"  # Pull from upstream
    # Push to ghcr.io/myorg/app
```

### Multi-Environment Deployments
```yaml
images:
  - image: "registry.io/app"
    target:
      registry: "prod-registry.io/apps"  # Production
  - image: "registry.io/app"
    target:
      registry: "staging-registry.io/apps"  # Staging
```


omercnet:
It might be sensible to split this to two PRs 
Let me know if you'd like that

pkg/cmd/cmd.go
```diff
@@ -131,6 +135,7 @@ copa patch --config copa-bulk-config.yaml --push (Bulk Image Patching)`,
 	}
 	flags := patchCmd.Flags()
 	flags.StringVar(&ua.configFile, "config", "", "Path to a bulk patch YAML config file (Comprehensive update only). Cannot be used with --image, --report, or --tag.")
+	flags.StringVar(&ua.patchedReportsDir, "patched-reports-dir", "", "Directory containing vulnerability reports for patched images (for skip detection in bulk mode with --config)")
```
robert-cronin: The help text says --patched-reports-dir is for bulk mode with --config, but I don’t see validation enforcing that; can we return an explicit error when this flag is set without --config instead of silently no-oping?

pkg/cmd/cmd.go
```diff
@@ -145,6 +150,7 @@ copa patch --config copa-bulk-config.yaml --push (Bulk Image Patching)`,
 	flags.DurationVar(&ua.timeout, "timeout", 5*time.Minute, "Timeout for the operation, defaults to '5m'")
 	flags.StringVarP(&ua.scanner, "scanner", "s", "trivy", "Scanner used to generate the report, defaults to 'trivy'")
 	flags.BoolVar(&ua.ignoreError, "ignore-errors", false, "Ignore errors and continue patching (for single-platform: continue with other packages; for multi-platform: continue with other platforms)")
+	flags.BoolVar(&ua.force, "force", false, "Force re-patching even if the patched image has no fixable vulnerabilities (bulk mode with --config only)")
```
robert-cronin: Same question for --force: the flag text says bulk-only, but it appears accepted in non-bulk mode with no effect; can we add validation for clearer UX?

pkg/bulk/engine.go
```diff
@@ -21,16 +22,59 @@ import (
...
+
+	// Parse the source image to extract the image name
+	ref, err := name.ParseReference(sourceImage)
+	if err != nil {
+		return "", fmt.Errorf("failed to parse source image '%s': %w", sourceImage, err)
+	}
+
+	// Extract the image name (last segment of the repository path)
+	repoStr := ref.Context().RepositoryStr()
+	repoParts := strings.Split(repoStr, "/")
```
robert-cronin, line 73: buildTargetRepository keeps only the last source path segment, so different sources like quay.io/team-a/redis and ghcr.io/team-b/redis both map to <target>/redis; should we preserve more path context or make mapping strategy configurable?

pkg/bulk/engine.go
```diff
@@ -119,11 +190,38 @@ func PatchFromConfig(ctx context.Context, configPath string, opts *types.Options
...
+				// Evaluate whether patching is needed and resolve the final tag
+				// Use targetRepo for skip detection (queries the registry where patched images are pushed)
+				action := evaluatePatchAction(targetRepo, targetTag, opts.Scanner, opts.Force, reports, opts.PkgTypes, opts.LibraryPatchLevel)
+				if action.ShouldSkip {
+					// Record as skipped
+					mu.Lock()
+					results = append(results, patchJobStatus{
+						Name:    spec.Name,
+						Source:  imageWithTag,
+						Target:  action.ResolvedTag,
```
robert-cronin: Skipped results show Target as tag-only while patched results show full repo:tag; could we standardize skipped Target to full reference (targetRepo:tag) for consistent cross-registry summaries?

pkg/bulk/skip.go
```diff
@@ -0,0 +1,291 @@
...
+
+// buildReportIndex scans a directory for JSON report files, extracts ArtifactName
+// from each, normalizes references, and builds a lookup map.
+func buildReportIndex(reportsDir string) *reportIndex {
+	idx := &reportIndex{
+		refs: make(map[string]string),
+	}
+
+	// Read all files in the directory
+	entries, err := os.ReadDir(reportsDir)
```
robert-cronin, line 133: buildReportIndex uses os.ReadDir (top-level only), so nested report directories are ignored; should we support recursive discovery (WalkDir) or clearly document the flat-directory requirement?

robert-cronin (commented):
Thank you for the contribution! just a few questions

pkg/cmd/cmd.go
omercnet: Agreed, added validation to return an explicit error when `--patched-reports-dir` is used without `--config`.

pkg/cmd/cmd.go
omercnet: Agreed, same treatment -- added validation for `--force` alongside `--patched-reports-dir` in the same check.

pkg/bulk/engine.go
omercnet: Good point. Documented this limitation in the function comment. Per-image `target.registry` overrides in the config provide an escape hatch for collision scenarios. Happy to revisit with a configurable mapping strategy if users run into this in practice.

pkg/bulk/engine.go
omercnet: Good catch, standardized the skipped Target to show the full `repo:tag` reference for consistent cross-registry summaries.

pkg/bulk/skip.go
omercnet: Documented the flat-directory requirement in the function comment and added a debug log when subdirectories are encountered. Recursive discovery adds complexity around duplicate ArtifactNames -- happy to add it in a follow-up if needed.

ashnamehrotra:
@omercnet thanks for adding the skip functionality! I think we can eliminate the`--force` flag to avoid creating extra flags for the CLI.  In this case, copa would create the same image with the patched tag which wouldn't be accurate and we intend on copa to exit out in this scenario. Is there a specific use case for having the same image with the patched tagged?

ashnamehrotra:
@omercnet Also forgot to mention, can we use the existing `-r` flag for report dir instead of `--patched-reports-dir`?

omercnet:
sure thing @ashnamehrotra, done!

omercnet:
@ashnamehrotra unless there's anything else, would appreciate merging this

robert-cronin (approved):

robert-cronin:
Added #1483 to resolve the vuln-check error. Other two failures are transient.

pkg/cmd/cmd.go
```diff
@@ -130,9 +130,9 @@ copa patch --config copa-bulk-config.yaml --push (Bulk Image Patching)`,
 		},
 	}
 	flags := patchCmd.Flags()
-	flags.StringVar(&ua.configFile, "config", "", "Path to a bulk patch YAML config file (Comprehensive update only). Cannot be used with --image, --report, or --tag.")
+	flags.StringVar(&ua.configFile, "config", "", "Path to a bulk patch YAML config file (Comprehensive update only). Cannot be used with --image or --tag.")
 	flags.StringVarP(&ua.appImage, "image", "i", "", "Application image name and tag to patch")
-	flags.StringVarP(&ua.report, "report", "r", "", "Vulnerability report file or directory path")
+	flags.StringVarP(&ua.report, "report", "r", "", "Vulnerability report file (single-image mode) or directory of reports for patched images (bulk mode)")
```
ashnamehrotra: small nit - the directory of reports isn't just for bulk patching, it is also for multi-platform patching. I think we can leave it at "Vulnerability report file or directory path"

omercnet:
```suggestion
	flags.StringVarP(&ua.report, "report", "r", "", "Vulnerability report file or directory of reports")
```

ashnamehrotra:
@omercnet did some local testing and seeing this bug:

- discoverExistingPatchTags() treats the architecture-suffix tag 3.18.0-patched-386 as version number 386 (the 386 arch is purely numeric)
- This causes latestPatchTag to return 3.18.0-patched-386 instead of 3.18.0-patched
- The report lookup then fails (no report for localhost:5111/alpine:3.18.0-patched-386)
- Skip detection falls through to fail-open and patches again as 3.18.0-patched-387
- The per-architecture tags (-arm64, -amd64, -386, etc.) are pushed by Copa's multi-platform patching flow, and this PR's version-suffix regex doesn't account for that.

omercnet:
great catch @ashnamehrotra ! added a fix

omercnet:
@ashnamehrotra @robert-cronin merge please..?

PR #1505, opened 2026-03-23
fix: resolve TUI freeze and CLI deadlock on early build errors

### LINKED ISSUE #1504: [BUG] Terminal freezes in raw mode and deadlocks on early patch errors
### Version of copa

v0.13.0-39-g59f49a7f

### Expected Behavior

The CLI should print the error, cleanly shut down the progrock TUI, restore the terminal to its normal state, and exit the program gracefully.

https://github.com/user-attachments/assets/48808fda-1a77-4297-b0d4-8f29fea1d16b

### Actual Behavior

The CLI prints the error but leaves the terminal hijacked by the TUI's raw mode and deadlocks the background worker threads.

https://github.com/user-attachments/assets/5018900f-831d-41fa-b148-de9965e4857d

### Steps To Reproduce

1. Use a vulnerability report missing necessary OS metadata e.g.
```json
{
	"SchemaVersion": 2,
	"ArtifactName": "docker.io/archlinux:base-devel-20250907.0.417472",
	"ArtifactType": "container_image",
	"Results": [
		{
			"Target": "archlinux (archlinux 20250907.0.417472)",
			"Class": "os-pkgs",
			"Type": "archlinux",
			"Vulnerabilities": [
				{
					"VulnerabilityID": "CVE-TEST-1234",
					"PkgName": "filesystem",
					"InstalledVersion": "2024.01.01",
					"FixedVersion": "2099.12.31",
					"Layer": {
						"DiffID": "sha256:dummylayer"
					},
					"Severity": "HIGH"
				}
			]
		}
	]
}
```

2. Run a patch command 
```
./copacetic patch \
  --image docker.io/archlinux:base-devel-20250907.0.417472 \
  --report arch_report.json \
  --tag arch-patched:test
```

3. See the error freeze the terminal

parthivsaikia:
**Remove** the `ch chan error` parameter from `patchWithContext`, `patchSingleArchImage`, and `executePatchBuild` to prevent short-circuiting the orchestration `errgroup`.
**Refactor** the orchestration logic to rely on standard `return err` bubbling. This ensures the main `select` block waits for `eg.Wait()` to finish, allowing the `progrock` TUI to execute its `defer term.Restore()` cleanup and preventing raw-mode terminal freezes.
**Add** `defer pipeW.Close()` to the BuildKit worker goroutine in `patchSingleArchImage` to guarantee the image loader stream always receives an EOF on early failures, preventing a permanent CLI deadlock.
**Remove** `trySendError(ch, err)` in the VEX document validation block in favor of standard error returns.

### Demonstrations

**Before (Terminal Freeze & Deadlock):**
https://github.com/user-attachments/assets/d5f4d2d1-619b-422d-ad3f-df7588ce4ca8

**After (Clean Exit & Terminal Restored):**
https://github.com/user-attachments/assets/caf34ffc-5e7c-4d6f-837f-3395378962f8

Closes #1504

### DIFF
--- a/integration/singlearch/patch_test.go
+++ b/integration/singlearch/patch_test.go
@@ -283,7 +283,7 @@ func patch(t *testing.T, ref, patchedTag, path string, ignoreErrors bool, report
 	out, err := cmd.CombinedOutput()
 
 	if strings.Contains(ref, "oracle") && reportFile && !ignoreErrors {
-		assert.Contains(t, string(out), "Error: detected Oracle image passed in\n"+
+		assert.Contains(t, string(out), "detected Oracle image passed in\n"+
 			"Please read https://project-copacetic.github.io/copacetic/website/troubleshooting before patching your Oracle image")
 	} else {
 		require.NoError(t, err, string(out))


pkg/patch/single.go
```diff
@@ -208,7 +207,8 @@ func patchSingleArchImage(
 	// Start the main build process and capture preserved states
 	var patchResult *Result
 	eg.Go(func() error {
-		result, err := executePatchBuild(ctx, ch, bkClient, buildConfig, buildkitImageRef, &targetPlatform,
+		defer pipeW.Close()
```
ashnamehrotra, line 210: maybe we can use `CloseWithError` here to allow the goroutine to receive the actual error

robert-cronin:
I think the CI failures are all the same thing. we just need to update the tests to expect the new error message format (i.e. `"Error: Patch Failed\n  failed to solve: detected Oracle image..."` instead of `"Error: detected Oracle image..."`).

ashnamehrotra:
@parthivsaikia I am not able to repro this on my terminal with the main branch. @robert-cronin would you be able to check if you can repro

robert-cronin:
> @parthivsaikia I am not able to repro this on my terminal with the main branch. @robert-cronin would you be able to check if you can repro

Checking against main, I was able to replicate the original bug, however it presented for me as escape characters that presumably were causing the original freezing:
image
and confirmed fix does not produce these escape characters:
image

robert-cronin:
There are a few PR failures. The vuln check will be solved by  #1512

PR #1516, opened 2026-04-02
fix: go patching log levels

robert-cronin:
Adds add ingress-nginx e2e test for trimpath/OCI label rebuild
Improve Go patching logging, error handling, and OCI label fallback for trimpath binaries


sozercan:
When Copa rebuilds Go binaries, vulnerability scanners see a pseudo-version like v0.0.0-20251103154607-ad643baabbbf instead of the real version (e.g., v1.14.0), which can get flagged.

example:
image

robert-cronin:
The vuln-check issue will be solved by https://github.com/project-copacetic/copacetic/pull/1525

pkg/provenance/rebuilder.go
```diff
@@ -76,11 +124,9 @@ if [ -n "$FIRST_DIR" ]; then
...
 fi
 
-# Nothing found - create placeholder so LLB does not fail
-echo "SKIP: no buildable Go source found for $BINARY_NAME"
-mkdir -p "$(dirname "$OUTPUT")"
-printf '#!/bin/sh\necho placeholder\n' > "$OUTPUT"
-chmod +x "$OUTPUT"
+# Nothing found
+echo "ERROR: no buildable Go source found for $BINARY_NAME"
+exit 1
```
omercnet, line 129: **Breaking behavioral change**: The old code created a placeholder binary so the LLB build could continue, allowing partial patches on multi-binary images. This `exit 1` will fail the entire build if *any* binary lacks source.

For users who currently get partial patches (some binaries rebuilt, others left as-is), this is a regression. Was this intentional?

If so, it might warrant a note in the PR description and potentially a deprecation/migration notice. If not, consider keeping the placeholder behavior (or making it configurable).

pkg/provenance/rebuilder.go
```diff
@@ -23,6 +23,54 @@ OUTPUT="$2"
...
+                commit|build|gitcommit|sha|revision|gitsha|commit_sha) val="$OCI_REVISION" ;;
+                repo|repository|source|url|repo_info) val="$OCI_SOURCE" ;;
+            esac
+            [ -n "$val" ] && XFLAGS="$XFLAGS -X $pkg.$varname=$val"
+        done
+    done
+
+    if [ -n "$XFLAGS" ]; then
+        echo "Copa: injecting version ldflags from OCI labels:$XFLAGS"
+        BUILD_PREFIX="$BUILD_PREFIX '-ldflags=$XFLAGS'"
```
omercnet, line 68: **Shell injection concern**: OCI label values flow from `formatOCILabelsForScript` → `/tmp/copa_oci_labels` → here, where they get interpolated into `BUILD_PREFIX` without sanitization.

A malicious image could set `org.opencontainers.image.revision` to something like `"; rm -rf /; #` and it would end up in the shell variable. The `while IFS=... read` loop above does not validate the values.

Consider either:
- Validating label values match expected patterns (hex for revision, URL for source, semver for version) in `formatOCILabelsForScript`
- Or shell-escaping values when writing to the labels file

pkg/langmgr/golang.go
```diff
@@ -17,6 +18,31 @@ import (
...
+// OCI annotation keys for source provenance (https://github.com/opencontainers/image-spec/blob/main/annotations.md).
+const (
+	ociAnnotationSource   = "org.opencontainers.image.source"
+	ociAnnotationRevision = "org.opencontainers.image.revision"
+)
+
+// extractSourceFromImageConfig reads OCI standard labels from the image config
+// to recover source repository and revision when VCS info is missing from the binary
+// (e.g. built with -trimpath or -buildvcs=false).
+func extractSourceFromImageConfig(configData []byte) (sourceURL, revision string) {
```
omercnet: **Duplicate JSON parsing**: `extractLabelsFromConfig()` in `buildkit.go` already parses this same image config JSON and stores labels in `config.ImageLabels`. This function re-parses the same raw bytes with a nearly identical struct definition.

Could this just read from the already-extracted `ImageLabels` map instead? That would eliminate the duplicate parsing and keep a single source of truth for label extraction.

pkg/langmgr/golang.go
```diff
@@ -416,12 +442,20 @@ func (gm *golangManager) upgradePackages(
...
 	rebuiltState, rebuildFailedPkgs, rebuildErr := gm.attemptBinaryRebuild(ctx, currentState, updates, stdlibFixedVersion)
 	if rebuildErr == nil {
-		log.Info("Successfully rebuilt Go binaries with updated dependencies")
+		log.Debug("Go binary rebuild LLB graph constructed successfully")
 		return rebuiltState, rebuildFailedPkgs, nil
 	}
+
+	// If the only updates are stdlib vulns (which require binary rebuild), don't
+	// fall back to go.mod updates — they can't fix compiled binaries.
+	if len(updates) == 0 && stdlibFixedVersion != "" {
```
omercnet, line 438: **Subtle logic concern**: This conflates "only stdlib updates exist" with "should not attempt go.mod fallback." If the rebuild failed for an unrelated reason (network issue, OOM, missing toolchain) but the update set *happens* to be stdlib-only, this prevents the fallback path entirely.

Was the intent to skip the fallback only when stdlib patches *cannot* be applied via go.mod? If so, the condition might need to also check the specific failure reason.

pkg/langmgr/golang.go
```diff
@@ -665,10 +722,16 @@ func (gm *golangManager) attemptBinaryRebuild(
...
-	if totalRebuilt < len(binaries) {
-		log.Warnf("Partial patch: %d/%d binaries rebuilt successfully. Failed: %v", totalRebuilt, len(binaries), rebuildErrors)
+	if totalRebuilt < totalAttempted {
+		log.Warnf("Partial patch: %d/%d attempted binaries rebuilt. Failed: %v", totalRebuilt, totalAttempted, rebuildErrors)
+		for _, errStr := range rebuildErrors {
+			// Extract the module name from error strings like "path: error"
+			if parts := strings.SplitN(errStr, ":", 2); len(parts) > 0 {
+				failedPackages = append(failedPackages, parts[0])
+			}
+		}
```
omercnet, line 725: **Fragile string parsing**: Extracting module names from error message strings via `SplitN(errStr, ":", 2)` will silently produce wrong package names if the error format ever changes (e.g., if a path contains `:`, or if the format is refactored).

Consider tracking failures structurally. This loop:

```suggestion
		for _, f := range rebuildFailures {
			failedPackages = append(failedPackages, f.binaryPath)
		}
```

…paired with a struct and changes at the recording sites:

```go
// Replace: var rebuildErrors []string
// With:
type rebuildFailure struct {
    binaryPath string
    reason     string
}
var rebuildFailures []rebuildFailure

// At each failure site, replace:
//   rebuildErrors = append(rebuildErrors, fmt.Sprintf("%s: no source commit available", binaryPath))
// With:
//   rebuildFailures = append(rebuildFailures, rebuildFailure{binaryPath, "no source commit available"})
```

This also gives you structured data for richer output (JSON, per-binary exit codes) without reverse-engineering formatted strings.

omercnet (changes requested):
The log level cleanup (Info→Debug) is solid — makes the default `copa patch` output much more focused. The OCI label fallback for trimpath binaries is a good direction for images like ingress-nginx.

Requesting changes on two items:

1. **Placeholder → `exit 1` in the rebuild script is a silent breaking change** for users who currently get partial patches on multi-binary images. This needs either a revert to placeholder behavior, a flag to opt in, or at minimum explicit documentation in the PR description.

2. **OCI label values flow unsanitized into shell variables** in `inject_version_ldflags`. Label values should be validated or escaped before being interpolated into `BUILD_PREFIX`.

The remaining comments (duplicate JSON parsing, string-based error tracking, stdlib short-circuit logic) are suggestions — happy to discuss.

Also noting CI has `Test Go patching` red, which is directly relevant to these changes.

pkg/provenance/rebuilder.go
robert-cronin: Thanks for flagging. Keeping `exit 1` intentionally. The partial-patch capability you mention is preserved at the caller level in `golang.go:682-720` (error accumulation plus "Partial patch: N/M rebuilt" warning). What actually changed is that a binary with no discoverable source now produces a *visible* error instead of silently writing a placeholder that looks patched but isn't. The old behavior was a foot-gun for users who'd trust the output. This isn't a regression in the partial-patch flow, just removal of the silent-success mode.

pkg/provenance/rebuilder.go
robert-cronin: Addressed in `c005d024`. Added allow-list validators `isSafeOCIVersion`, `isSafeOCIRevision`, `isSafeOCISource` in `formatOCILabelsForScript` (rebuilder.go:148-230). Values that fail validation are skipped with a warning rather than interpolated.

pkg/langmgr/golang.go
robert-cronin: Addressed in `c005d024`. `extractLabelsFromConfig` is now only called once in `buildkit.go:138`, and golang.go reads from `gm.config.ImageLabels` (see golang.go:597, 645-647, 683). Removed the duplicate struct and parse.

pkg/langmgr/golang.go
robert-cronin: Keeping the current logic. The `go.mod` fallback genuinely cannot fix stdlib vulns, since stdlib is part of the toolchain, not go.mod. When updates are stdlib-only and rebuild failed (for any reason, including network/OOM), attempting the fallback is pointless; we would do work that produces no patched code, then still fail. The current short-circuit surfaces the real error rather than masking it behind a misleading fallback. Happy to improve the log message to make the reason clearer if you think that would help.

pkg/langmgr/golang.go
robert-cronin: Agreed the struct would be cleaner. Holding off on it in this PR to keep scope tight, will track as a follow-up issue. The `:` format is entirely controlled by our own `fmt.Sprintf` calls in this file (not parsing external input), so the fragility is bounded, and a refactor touches 6+ call sites.

robert-cronin: https://github.com/project-copacetic/copacetic/issues/1559

ashnamehrotra (approved):

omercnet (approved):

PR #1517, opened 2026-04-04
feat: add patch summary output showing total/patched/skipped vulnerabilities

### LINKED ISSUE #1451: [REQ] Add patch summary output showing total/patched/skipped vulnerabilities

north-echo:
## Summary
- Display a patch summary (total vulnerabilities considered, patched, skipped) after `copa patch` completes
- Summary is computed inline during report parsing using the same filtering and patch-level logic as the actual patch, ensuring counts always match
- Summary generation is non-fatal: if unavailable (e.g. custom scanner or no report), it is silently omitted

## Motivation

Issue #1451 — Copa currently outputs only the patched image reference after patching, giving users no visibility into what was patched vs skipped. Users must re-scan to understand results.

This is an alternative to #1468 that incorporates the review feedback from that PR:

| Feedback | How addressed |
|---|---|
| Summary error aborts entire patch (robert-cronin) | Summary is a nil-safe pointer; no new error paths |
| Report parsed twice (Copilot) | Single parse — counts computed inline during `ParseWithLibraryPatchLevel` |
| Summary ignores `libraryPatchLevel` (Copilot) | Counts use identical patch-level logic; a vuln whose fix is excluded by `--library-patch-level` counts as skipped |
| "Total in report" misleading with `--pkg-types` filter (Copilot) | Label says "Total vulnerabilities considered"; counts respect `--pkg-types` filtering |
| Library vulns overcounted as patched (Copilot) | Vulns where `FindOptimalFixedVersionWithPatchLevel` returns "" count as skipped |
| Should work without a report (ashnamehrotra) | Deferred — requires plumbing counts out of BuildKit (documented as follow-up) |

## Changes

- **`pkg/types/unversioned/types.go`** — Add `PatchSummary` struct; add `OSSummary`/`LibrarySummary` fields and `CombinedSummary()` method on `UpdateManifest`
- **`pkg/report/trivy.go`** — Count vulnerabilities inline during `ParseWithLibraryPatchLevel` (OS and library tracked separately for `--pkg-types` correctness)
- **`pkg/report/report.go`** — Nil out irrelevant summary half during `pkgTypes` filtering
- **`pkg/types/types.go`** — Add `Summary` field to `PatchResult`
- **`pkg/patch/single.go`** — Attach summary to `PatchResult` on both success and `ErrNoUpdatesFound` paths
- **`pkg/patch/patch.go`** — Add `logPatchSummary` helper; print summary before "Patched image" line at all call sites

## Example output

```
Patch Summary:
  Total vulnerabilities considered: 24
  Patched:                           6
  Skipped (no fix available):       18

Patched image (linux/amd64): docker.io/example/image:patched
```

## Known limitations (follow-up)
- No summary in no-report (comprehensive update) flows
- Multi-platform images print summary per-arch without platform label
- "Patched" counts CVEs not unique packages

## Test plan
- [x] Unit tests for mixed report, all fixable, none fixable, empty report
- [x] Unit test for patch-level skipping (major bump under `--library-patch-level=patch`)
- [x] Unit test for `--pkg-types` filtering
- [x] Unit test for .NET unpatchable package skipping
- [x] Unit tests for `CombinedSummary()` (both, one, neither)
- [x] Existing `TestTryParseScanReport` updated for new summary field
- [x] `go test ./pkg/...` passes (excluding pre-existing buildkit/utils env failures)
- [x] `golangci-lint` clean
- [x] `gofumpt` clean

Closes #1451

Assisted-by: Claude (Anthropic)


robert-cronin (approved):
The counting logic looks correct to me. OS and library vulns are tracked separately so `--pkg-types` filtering can nil out the irrelevant half, and the Total = Patched + Skipped invariant holds across all paths (no-fix, patch-level-blocked, and unpatchable .NET packages). Test coverage is thorough.

One small thing: the summary alignment in `logPatchSummary` uses fixed-width spacing which will look off if the logrus prefix length varies between runs. Not a blocker, just something that might look a bit odd in practice.

north-echo:
Thanks for the review! Good catch on the alignment — collapsed the summary into a single log line to avoid the variable-prefix issue.

pkg/report/trivy.go
```diff
@@ -525,6 +542,20 @@ func (t *TrivyParser) ParseWithLibraryPatchLevel(file, libraryPatchLevel string)
 			}
 
 			optimalVersion := FindOptimalFixedVersionWithPatchLevel(info.InstalledVersion, fixedVersions, patchLevelToUse)
+
+			// Count CVEs in this package group for the library summary.
+			// If patch-level constraints prevent patching (optimalVersion == ""),
+			// these vulns had a fix but it couldn't be applied — count as skipped.
+			vulnCount := len(langPackageVulnIDs[key])
+			if vulnCount == 0 {
```
omercnet, line 550: **Nit**: When would a package group have zero tracked CVE IDs but still reach this point? Is this a defensive guard, or can this actually happen in practice? If defensive, a short comment explaining the edge case would help future readers.

omercnet (approved):
Clean implementation. The inline counting during `ParseWithLibraryPatchLevel` is the right call — it guarantees the summary matches the actual patch decisions. Supersedes #1468 with all review feedback addressed.

PR #1526, opened 2026-04-08
fix: prevent Go module flag injection via leading-dash names

sozercan:
### Motivation
- The Go patching paths build `go get` shell commands from report-provided module names but only validated for shell metacharacters and whitespace, allowing module names that start with `-` to be interpreted as `go` flags and enabling argument injection. 
- The fix must block leading-dash module/package names in all update code paths while preserving existing behavior for valid module names.

### Description
- Add a leading-dash rejection to `validateGoPackageName` by returning an error if a module name `strings.HasPrefix(name, "-")`.
- Enforce the same guard in `updateGoModule` and `upgradePackagesWithTooling` by rejecting `u.Name` values that start with `-` before constructing `go get` commands.
- Introduce `validateGoModuleName(module string)` in `pkg/provenance/rebuilder.go` and use it when applying updates to centralize validation and reject leading-dash module names.
- Add regression tests: extend `TestValidateGoPackageName` with a leading-dash case and add `TestValidateGoModuleName` in rebuilder tests to cover unsafe characters and leading-dash rejection.

### Testing
- Ran unit tests with `go test ./pkg/langmgr ./pkg/provenance` and they passed successfully. 
- New/updated tests include `TestValidateGoPackageName` and `TestValidateGoModuleName`, and both passed under the test run.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69d682c378d4832ab3c18d94c6e70a02)

### DIFF
--- a/pkg/langmgr/golang.go
+++ b/pkg/langmgr/golang.go
@@ -63,6 +63,11 @@ func validateGoPackageName(name string) error {
 		return fmt.Errorf("package name contains whitespace: %s", name)
 	}
 
+	// Module names must not start with '-' to prevent go tool flag injection.
+	if strings.HasPrefix(name, "-") {
+		return fmt.Errorf("package name cannot start with '-': %s", name)
+	}
+
 	return nil
 }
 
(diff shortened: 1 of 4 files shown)


robert-cronin (approved):
Good security fix, clean extraction of validateGoModuleName. LGTM

PR #1532, opened 2026-04-08
fix: restore strict multi-platform failure behavior when ignore-errors is false

sozercan:
### Motivation
- The multi-platform patcher recorded per-platform errors but every goroutine returned `nil`, so `errgroup` did not surface failures and a run with `--ignore-errors=false` could succeed despite partial platform failures.

### Description
- Add a post-`g.Wait()` check in `pkg/patch/multi.go` to return an error when `hasErrors && !ignoreError`, while retaining the existing behavior that only fails when all real patch attempts failed when `ignoreError` is true.

### Testing
- Ran `go test ./pkg/patch`, which attempted to execute unit tests but the run failed due to external module download being blocked (`proxy.golang.org` returned HTTP 403), so tests could not complete in this environment.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69d6c929e614832a82b71d84eac83799)


pkg/patch/multi.go
```diff
@@ -298,7 +298,12 @@ func patchMultiPlatformImage(
 	// Wait for the progress display to finish
 	_ = displayEg.Wait()
 
-	// Check if we should fail based on the results:
+	// With strict error handling, fail if any platform encountered an error.
+	if hasErrors && !ignoreError {
+		return fmt.Errorf("one or more platform patches failed, see summary for details")
+	}
```
omercnet, line 328: **Summary rendering order**: This return fires before the `tui.RenderPatchSummary()` block further down, so "see summary for details" is misleading - users will see the error but no summary output.

The existing all-failed check on line 308 has the same issue.

Consider moving the summary rendering above both checks so it prints before returning. The per-platform breakdown is useful context for debugging which platform failed and why.

omercnet (approved):

robert-cronin (approved):
LGTM

robert-cronin:
The dotnet failure looks like a bug unrelated to this change, we can fix it in another PR

PR #1533, opened 2026-04-08
fix(nodejs): validate npm tarballs before extraction

sozercan:
### Motivation

- The code previously extracted npm tarballs directly into `node_modules` using `tar xzf` without validating archive members, exposing a zip-slip / path traversal risk when processing untrusted tarballs. 
- Add a minimal defence-in-depth fix that validates tarball contents before extraction while preserving the existing direct-tarball replacement strategy.

### Description

- Added a `safeTarExtractScript` helper that inspects a tarball, rejects absolute or `..` traversal member names and symlink/hardlink entries, and then extracts with `--no-same-owner --no-same-permissions` and `--strip-components=1` into the target directory.
- Replaced all direct `tar xzf ... -C <dir>` invocations in Node.js package replacement flows (direct and transitive updates, app-local, global packages, and npm-internal replacements) to invoke the new validated extraction script instead of raw extraction.
- Changes are localized to `pkg/langmgr/nodejs.go` and preserve existing post-extract checks (e.g., verifying `package.json` exists after extraction).

### Testing

- Ran `gofmt -w pkg/langmgr/nodejs.go` to format changes successfully.
- Ran `go test ./pkg/langmgr -run TestNode -count=1`, which passed (`ok github.com/project-copacetic/copacetic/pkg/langmgr`).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69d6cabed4e8832ab09e3c4bf8590802)


robert-cronin (approved):
Copilot already flagged the main things (shell quoting, test coverage, awk/BusyBox portability). I think the shell-quoting and unit test suggestions are good ones to include.

On the awk and `--no-same-owner` portability points, I think this runs inside the BuildKit tooling container rather than the target image, so those might be less of a concern. Worth double-checking though.

PR #1535, opened 2026-04-08
fix(ci): prevent tag-based command injection in release workflow

sozercan:
### Motivation
- The release workflow interpolated attacker-controlled tag/ref values directly inside shell `run:` blocks, which allows command substitution when a tag contains `$()` or backticks. 
- The goal is to eliminate the command-injection sink while preserving existing release/version behavior.

### Description
- Hardened `.github/workflows/release.yml` by moving GitHub expression values into step-level `env:` variables and referencing those shell variables inside `run:` blocks instead of embedding `${{ ... }}` directly. 
- Replaced direct use of `${{ github.ref }}` with `RAW_GITHUB_REF` and used safe shell parameter expansion (`${RAW_GITHUB_REF#refs/tags/v}`) to derive `VERSION_TAG_ONLY` inside the step. 
- Replaced direct interpolation of `${{ steps.vars.outputs.version_tag }}` and `${{ github.repository_owner }}` inside run scripts with `VERSION_TAG_ONLY`, `RELEASE_TAG`, and `IMAGE_OWNER` step env variables. 
- Applied the same pattern to all image build and release-branch creation steps so no attacker-controlled GitHub expressions are parsed by the shell at runtime. 

### Testing
- Ran `sed -n '1,140p' .github/workflows/release.yml` to inspect the updated step body and it completed successfully. 
- Verified no direct interpolation remains by running `rg -n "\$\{\{ github\.ref \}\}|\$\{\{ steps\.vars\.outputs\.version_tag \}\}|\$\{\{ github\.repository_owner \}\}" .github/workflows/release.yml` and observed no matches. 
- Reviewed the `git diff` of `.github/workflows/release.yml` to confirm the changes were applied and the workflow behavior for version/tag extraction is unchanged.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69d6c90da9f4832aaf90ce0d02f9cf98)


.github/workflows/release.yml
```diff
@@ -99,9 +103,12 @@ jobs:
...
       - name: Build and push copa-action image with new version
+        env:
+          RELEASE_TAG: ${{ steps.vars.outputs.version_tag }}
+          IMAGE_OWNER: ${{ github.repository_owner }}
         run: |
-          tag="${{ steps.vars.outputs.version_tag }}"
-          image_owner="${{ github.repository_owner }}"
+          tag="${RELEASE_TAG}"
+          image_owner="${IMAGE_OWNER}"
           docker buildx build --build-arg copa_version=${tag} -t "ghcr.io/${image_owner}/copa-action:v${tag}" --push ./copa-action
```
robert-cronin, line 112: While we are here, might as well quote `${tag}` for consistency:
```suggestion
          docker buildx build --build-arg copa_version="${tag}" -t "ghcr.io/${image_owner}/copa-action:v${tag}" --push ./copa-action
```

.github/workflows/release.yml
```diff
@@ -122,9 +129,12 @@ jobs:
...
       - name: Build and push copa-extension image with new version
+        env:
+          RELEASE_TAG: ${{ steps.vars.outputs.version_tag }}
+          IMAGE_OWNER: ${{ github.repository_owner }}
         run: |
-          tag="${{ steps.vars.outputs.version_tag }}"
-          image_owner="${{ github.repository_owner }}"
+          tag="${RELEASE_TAG}"
+          image_owner="${IMAGE_OWNER}"
           docker buildx build --push --platform linux/amd64,linux/arm64 --build-arg copa_version=${tag} -t "ghcr.io/${image_owner}/copa-extension:v${tag}" ./copa-extension/container/copa-extension
```
robert-cronin, line 138: Same here:
```suggestion
          docker buildx build --push --platform linux/amd64,linux/arm64 --build-arg copa_version="${tag}" -t "ghcr.io/${image_owner}/copa-extension:v${tag}" ./copa-extension/container/copa-extension
```

robert-cronin (approved):
Also worth noting that release-docs.yml has the same ${{ github.ref }} interpolation pattern in a run: block, could be a good follow-up.

PR #1537, opened 2026-04-09
fix(dotnet): validate .NET deps.json script inputs to prevent command injection

sozercan:
### Motivation
- The `buildUpdateDepsJsonScript` function interpolated untrusted `Name`/`InstalledVersion`/`FixedVersion` values directly into a `sh -c` script, creating a command-injection risk when processing malicious reports. 
- The change aims to eliminate the injection vector while preserving existing deps.json update behavior for valid updates.

### Description
- Filter inputs before script generation by validating updates into a `safeUpdates` slice using the existing `validateDotnetPackageName` and `validateDotnetVersion` helpers and skip any updates that fail validation. 
- Return a safe no-op script (`echo "No valid updates to apply to deps.json"`) when no validated updates remain. 
- Use `safeUpdates` instead of raw `updates` when building the jq/shell script so only validated values are interpolated. 
- Add two unit tests in `pkg/langmgr/dotnet_test.go`: `TestBuildUpdateDepsJsonScript_FiltersUnsafeUpdates` and `TestBuildUpdateDepsJsonScript_NoValidUpdates` to assert unsafe payloads are excluded and that all-invalid input yields the no-op script.

### Testing
- Added unit tests in `pkg/langmgr/dotnet_test.go` covering filtering behavior and the no-op outcome for all-invalid updates. 
- Attempted `go test ./pkg/langmgr`, but test execution failed in this environment due to blocked module downloads from `proxy.golang.org` (dependency resolution returned `Forbidden`), so automated tests could not be completed here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69d731977fcc832a9c27edb13e60943c)


robert-cronin (approved):
I think the Copilot suggestions here are worth including, especially the grep -F fix and the prerelease version mismatch concern.

sozercan:
@codex can you handle the grep -F fix and the prerelease version mismatch

PR #1538, opened 2026-04-09
fix(nodejs): prevent shell injection via untrusted package paths

sozercan:
### Motivation
- The Node.js patching flow derived application directories from untrusted `PkgPath` values and interpolated them directly into `sh -c 'cd %s && ...'` commands, creating a command-injection vector. 
- The fix must be minimal and preserve existing patching behavior while preventing shell metacharacters from breaking out of the scripted shell context.

### Description
- Add `shellQuote(s string) string` to escape embedded single quotes and produce a safely quoted shell argument. 
- Change vulnerable `sh -c` usages to pass the directory as a positional parameter (`$1`) via `sh -c 'cd -- "$1" && ...' -- %s` and supply the quoted directory using `shellQuote(...)` to avoid direct interpolation. 
- Apply the safer pattern across app-level flows and global-package flows where `workDir`/`pkgPath` was previously interpolated (devDependency removal, direct and transitive replacement commands, and cleanup steps). 
- Add `TestShellQuote` unit tests to validate quoting behavior for typical and single-quote-containing paths.

### Testing
- Ran unit tests for the modified package with `go test ./pkg/langmgr`, which passed (`ok github.com/project-copacetic/copacetic/pkg/langmgr`).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69d78bd22850832a9b40f359eabe161b)

### DIFF
--- a/pkg/langmgr/nodejs.go
+++ b/pkg/langmgr/nodejs.go
@@ -506,6 +506,12 @@ func (nm *nodejsManager) upgradePackages(
 	return &updatedState, nil
 }
 
+// shellQuote wraps s in single quotes and escapes embedded single quotes so it can be
+// safely passed as a shell argument.
+func shellQuote(s string) string {
+	return "'" + strings.ReplaceAll(s, "'", `'"'"'`) + "'"
+}
+
 // installNodePackages updates both direct and transitive dependencies.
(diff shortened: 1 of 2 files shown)


robert-cronin (approved):
LGTM

PR #1539, opened 2026-04-09
ci: block self-hosted build workflow jobs on forked PRs

sozercan:
### Motivation
- The build workflow used the self-hosted runner label `oracle-vm-16cpu-64gb-x86-64` while also triggering on `pull_request`, allowing untrusted forked PR code to execute on project-controlled infrastructure.
- Mitigate CI environment compromise risk by preventing fork-origin PRs from running on self-hosted runners while preserving existing behavior for pushes, `workflow_dispatch`, and same-repo PRs.

### Description
- Added a job-level guard to all jobs that run on the self-hosted runner to restrict execution to non-PR events or PRs originating from the same repository by inserting `if: ${{ github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository }}`.
- Applied the condition uniformly across the `jobs:` in `.github/workflows/build.yml` to cover unit, build, and all test jobs that previously used the `oracle-vm-16cpu-64gb-x86-64` label.
- This is a minimal, non-functional change to workflow logic that preserves existing CI behavior for trusted contexts and avoids running untrusted code on self-hosted infrastructure.

### Testing
- Ran `git diff --check -- .github/workflows/build.yml` to verify no whitespace/formatting issues and it returned clean (success).
- Confirmed the guard was inserted in all expected locations with `rg -n "full_name == github.repository" .github/workflows/build.yml` and observed `16` occurrences (success).
- Attempted to validate YAML with a Python parse but the environment lacked `PyYAML`, so automated YAML parsing was not performed (tooling missing in this environment).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69d6f82b04b4832a98251bc24390a49e)


robert-cronin (changes requested):
Copilot already flagged this but I want to emphasize it: `test.yml` has 7 jobs using the same `oracle-vm-16cpu-64gb-x86-64` self-hosted runner that are still unprotected (`build`, `test-gateway-all`, `test-gateway-alpine`, `test-gateway-debian`, `test-gateway-rpm`, `test-gateway-crossarch`, `test-gateway-layer-config`). I think the same guard needs to be applied there too, otherwise this only covers half the attack surface.

robert-cronin (approved):

PR #1540, opened 2026-04-09
fix: prevent apt option injection from distroless package names

sozercan:
### Motivation
- A distroless patching flow fed untrusted package names from `status.d` into `apt/apt-get` shell invocations unquoted, allowing apt option injection via leading dashes or embedded whitespace.  

### Description
- Validate parsed Debian package names in `GetPackageInfo` and reject names that don't match a safe Debian package-name regexp.  
- Harden dpkg DB regeneration by passing package names to `apt-get install` with `--` and quoting `apt show` arguments to prevent option parsing.  
- Produce `packages.txt` as newline-delimited entries and write it with `printf` to avoid word-splitting issues.  
- Update the embedded `apt_get_download.sh` to load `packages.txt` into a bash array and call `apt-get download -- ... "${packages[@]}"` to avoid unquoted expansion.  
- Modified files: `pkg/pkgmgr/dpkg.go`, `pkg/pkgmgr/scripts/apt_get_download.sh`.  

### Testing
- Ran `go test ./pkg/pkgmgr/...` and it passed (`ok github.com/project-copacetic/copacetic/pkg/pkgmgr`).  
- Ran `gofmt -w` on the modified Go file to ensure formatting consistency.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69d7abc76808832a867d33221c595e2e)


robert-cronin (changes requested):
The bash/sh incompatibility flagged by Copilot and Codex is a blocker. I think the POSIX rewrite approach (replacing `mapfile` with a `while read` loop) is the way to go rather than switching to bash, since the tooling image may not have bash available. The regex compilation and missing test suggestions are also worth addressing.

pkg/pkgmgr/dpkg.go
```diff
@@ -332,6 +335,11 @@ func GetPackageInfo(file string) (string, string, error) {
 	return packageName, packageVersion, nil
 }
 
+func isValidDebianPackageName(name string) bool {
+	// Debian package names are lowercase and may contain digits, plus, hyphen, and period.
+	return regexp.MustCompile(`^[a-z0-9][a-z0-9+.-]*$`).MatchString(name)
```
omercnet, line 340: **Regex recompilation per call**: `regexp.MustCompile` is called on every invocation. On images with a large status.d directory (hundreds of packages), this pays full compilation cost per package. Compile once at package scope:

```go
var debPkgNameRE = regexp.MustCompile(`^[a-z0-9][a-z0-9+.-]*$`)

func isValidDebianPackageName(name string) bool {
    return debPkgNameRE.MatchString(name)
}
```

pkg/pkgmgr/dpkg.go
```diff
@@ -320,6 +320,9 @@ func GetPackageInfo(file string) (string, string, error) {
 	} else {
 		return "", "", fmt.Errorf("no package name found for package")
 	}
+	if !isValidDebianPackageName(packageName) {
```
omercnet, line 323: **Missing test coverage**: This validation is security-critical but has no test cases. Suggested additions for dpkg_test.go:

- `"-malicious"` - reject (leading dash, option injection)
- `"--force-all"` - reject (double dash)
- `"foo bar"` - reject (space, word splitting)
- `"libssl1.1"` - accept
- `"g++"` - accept
- `"libc6.1"` - accept

robert-cronin (approved):

PR #1543, opened 2026-04-09
fix(generate): cap buffered patch layer size to mitigate memory DoS

sozercan:
### Motivation
- The `generate` flow buffered BuildKit tar exports into an unbounded `[]byte` which could be amplified by hardlink handling and cause high memory usage (DoS) for large or adversarial images. 
- Prevent unbounded in-memory buffering of the patch/diff layer and fail early when layer size exceeds a safe threshold.

### Description
- Introduce a new `maxPatchLayerSize` constant (1GB) and enforce it as an upper bound for the buffered patch layer. 
- Replace the raw `bytes.Buffer` sink used by BuildKit export with a `limitedBufferWriter` that enforces the `maxPatchLayerSize` during the export stream. 
- Add a defensive size check at the start of `createTarStream` to reject oversized `patchLayer` inputs. 
- Add unit tests: `TestLimitedBufferWriter_ExceedsLimit` and `TestCreateTarStream_PatchLayerTooLarge`, and keep existing `TestCreateTarStream` and `TestTarWriter` to ensure behavior is unchanged.

### Testing
- Ran targeted unit tests: `go test ./pkg/generate -run 'Test(LimitedBufferWriter_ExceedsLimit|CreateTarStream_PatchLayerTooLarge|CreateTarStream|TarWriter)' -count=1`, and they passed (`ok github.com/project-copacetic/copacetic/pkg/generate`).
- The new tests verify the capped writer behavior and that an oversized patch layer is rejected early.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69d756309b2c832aba2ca12f3bc3c4a2)


pkg/generate/generate.go
```diff
@@ -520,3 +528,26 @@ func (tw *tarWriter) Close() error {
...
+	limit   int64
+}
+
+func (w *limitedBufferWriter) Write(p []byte) (int, error) {
+	remaining := w.limit - w.written
+	if remaining <= 0 {
+		return 0, errors.Errorf("patch layer exceeds maximum allowed size of 1GB")
+	}
+
+	if int64(len(p)) > remaining {
```
robert-cronin, line 544: Writing partial data (`p[:remaining]`) then returning an error creates a corrupt tar in the buffer. I think it would be safer to return `(0, error)` without writing anything, so the buffer stays valid up to the point of failure.

robert-cronin (changes requested):
I think the OOM test and hardcoded "1GB" issues Copilot flagged are worth fixing. One additional thing:

robert-cronin (approved):

PR #1544, opened 2026-04-09
fix(ci): pin scanner-plugin-template dependency in build workflow

sozercan:
### Motivation
- The CI workflow installed `github.com/project-copacetic/scanner-plugin-template@latest`, which is an unpinned mutable dependency executed during CI and presents a supply-chain risk.

### Description
- Replace `go install github.com/project-copacetic/scanner-plugin-template@latest` with a pinned `@v0.0.1` in two places of `.github/workflows/build.yml` without changing other workflow logic.

### Testing
- Verified both occurrences now use the pinned version with `rg -n "scanner-plugin-template@" .github/workflows/build.yml`.
- Ensured no formatting or whitespace issues with `git diff --check`.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69d7bda52580832a9986b7ee8578a5b2)


robert-cronin (changes requested):
`v0.0.1` does not exist for `github.com/project-copacetic/scanner-plugin-template`, the repo has never been tagged. `go install ...@v0.0.1` will fail with `unknown revision v0.0.1` and break CI.

I think the options are:
1. Use the pseudo-version that `@latest` currently resolves to (e.g. `@v0.0.0-20240403201223-d6c20d3078c2`)
2. Create a proper `v0.0.1` tag in the scanner-plugin-template repo first, then use it here

Option 2 would probably be cleaner long term.

robert-cronin (approved):

PR #1525, opened 2026-04-08
fix: migrate docker/docker to moby/moby/client

robert-cronin:
Migrates production code from `github.com/docker/docker` to `github.com/moby/moby/client@v0.4.0` and `github.com/moby/moby/api@v1.54.1` to fix two Docker Engine vulnerabilities flagged by `govulncheck`:

- **GO-2026-4887**: AuthZ plugin bypass with oversized request bodies
- **GO-2026-4883**: Off-by-one error in plugin privilege validation

Changes:
- `pkg/imageloader/docker.go` — adapted to new `ImageLoadResult` (io.ReadCloser), `Ping(ctx, PingOptions)` signature, and `Opt` type
- `pkg/utils/mediatype.go` — import path swap
- Tests updated to match new API types

`docker/docker` remains as an indirect dependency used only by integration/e2e test infrastructure (not in the vulnerable code path).

PR #1546, opened 2026-04-11
feat: fallback source resolution for Go binary patching on stripped/distroless images

### LINKED ISSUE #1545: [REQ] Fallback source resolution for Go binary patching on stripped/distroless images

Go binary patching (#1388) clones source using the VCS commit hash from go version -m. When that metadata is stripped — common in distroless/scratch images built with -trimpath or -buildvcs=false — the clone fails and Copa falls back to a synthetic go.mod that often produces broken rebuilds. This affects roughly half of production Go images.

Two fallback strategies would cover most of these cases:

Image tag heuristic — extract the tag from the image ref (e.g. prometheus:v3.9.1 → v3.9.1), try it as a git ref against the repo derived from the module path. Works whenever image tags match git tags (common convention).

--go-vcs-url override — let users provide repo@ref explicitly for binaries where neither VCS metadata nor tag matching works.

Resolution order: user override → VCS commit (existing) → image tag heuristic.

Separately, deriveRepoFromModulePath could handle more vanity imports that currently return empty (cloud.google.com/go, go.uber.org, go.etcd.io, go.opentelemetry.io, google.golang.org/grpc, google.golang.org/protobuf).

omercnet:
## Summary

Resolves #1545

Enhances Go binary patching to work on distroless/scratch images where VCS metadata is stripped (`-trimpath` or `-buildvcs=false`). Adds a 4-step source resolution fallback chain, OCI source label handling, an OpenTelemetry lookup table, and better vanity import coverage.

### 4-step source resolution fallback (`pkg/provenance/rebuilder.go`)

`cloneSourceCode()` now tries four methods in priority order:

| Priority | Method | When it fires |
|---|---|---|
| 1 | `--go-vcs-url` override | User provides `repo@ref` explicitly (requires `COPA_EXPERIMENTAL=1`) |
| 2 | VCS commit from binary | Binary has embedded `vcs.revision` (existing behavior) |
| 3 | OCI source label | Image has `org.opencontainers.image.source` label on a trusted host, combined with a semver-like image tag |
| 4 | Image tag heuristic | No VCS info, no usable label, but image has a semver-like tag (e.g., `prometheus:v3.9.1` -> try `v3.9.1` as git ref) |

The OCI label step only fires when `validateRepoURL` passes; otherwise it falls through to the tag heuristic so users on GitLab/Gitea/self-hosted forges are not blocked.

New helpers:
- `parseGoVCSURL()` - parses `repo@ref` with `validateRepoURL` + `validateShellSafeStrict` for the ref
- `extractImageTag()` - strips digest suffix and registry-port prefix
- `looksLikeSemverTag()` - gates the tag heuristic so `latest`, `stable`, `alpine`, `sha-<hex>`, etc. do not trigger fallback
- `formatOCILabelsForScript()` + `isSafeOCIVersion` / `isSafeOCIRevision` / `isSafeOCISource` - safely propagate OCI labels into the rebuild script

### Extended vanity import mappings (`deriveRepoFromModulePath`)

Added resolution for vanity import paths that previously returned empty:
- `cloud.google.com/go/*` -> `github.com/googleapis/google-cloud-go`
- `go.uber.org/*` -> `github.com/uber-go/*`
- `go.etcd.io/*` -> `github.com/etcd-io/*`
- `go.opentelemetry.io/*` -> looked up via `otelRepos` map (`otel` -> `opentelemetry-go`, `contrib` -> `opentelemetry-go-contrib`, `collector` -> `opentelemetry-collector`, `ebpf-profiler` -> `opentelemetry-ebpf-profiler`); unknown top-level modules fall through to generic derivation
- `google.golang.org/grpc` -> `github.com/grpc/grpc-go`
- `google.golang.org/protobuf` -> `github.com/protocolbuffers/protobuf-go`

### `--go-vcs-url` experimental flag (`pkg/cmd/cmd.go`)

New flag behind `COPA_EXPERIMENTAL=1`:
```bash
copa patch -i calico/node:v3.31.4 --pkg-types library \
    --go-vcs-url https://github.com/projectcalico/calico@v3.31.4
```

### Synthetic binary info for detection failures

`collectGoBinaryInfo()` pulls binary paths and Go version from Trivy report `PkgPath` when `go version -m` detection fails (distroless/scratch). `buildSyntheticBinaryInfo()` constructs `BinaryInfo` entries from the collected paths so the rebuild flow can still fire.

### Rebuild verification

After constructing the LLB graph, `RebuildBinary` does an intermediate `Solve` and checks the rebuilt binary with `go version -m` before claiming success. Prevents the "LLB graph built, actual build fails later" class of bugs.

### Pipeline threading

`ImageRef`, `GoVCSURL`, and `ImageSourceLabel` threaded through: `cmd.go` -> `Options` -> `patch/core.go` -> `patch/single.go` -> `langmgr.go` -> `golangManager` -> `RebuildContext` -> `cloneSourceCode()`. `ImageLabels` (added by #1516) is also carried on `RebuildContext` and reused for OCI label propagation.

### Documentation (`website/docs/app-level-patching.md`)

- "How Binary Rebuilding Works" updated for the 4-step resolution chain
- "Source Resolution Chain" section describes each step with its preconditions and logs
- `--go-vcs-url` examples carry explicit `COPA_EXPERIMENTAL=1` callouts
- Limitations updated to reflect the new fallbacks

## Tests

- `TestExtractImageTag` - registry with port, digest-only, tag+digest, no tag
- `TestParseGoVCSURL` - tag ref, commit ref, missing ref, missing separator, untrusted host, unsafe ref
- `TestDeriveRepoFromModulePath` - extended with vanity import cases including OpenTelemetry lookup
- `TestCloneSourceCodeFallbackChain` - priority order (override > VCS > OCI label > tag > fail)
- `TestFormatOCILabelsForScript` - safe propagation of OCI labels into shell script
- `TestIsSafeOCIVersion` / `TestIsSafeOCIRevision` / `TestIsSafeOCISource` - injection-safe validators for OCI label values
- Updated `GetLanguageManagers` call sites in `dotnet_test.go`, `golang_test.go`, `langmgr_test.go`, `python_test.go`

## Verification

- `go build ./...` clean
- `go vet ./...` clean
- `gofmt -l` clean
- `golangci-lint run --new-from-rev=origin/main` zero new issues
- `go test ./pkg/langmgr/... ./pkg/provenance/... ./pkg/report/...` pass

### DIFF
--- a/pkg/langmgr/golang.go
+++ b/pkg/langmgr/golang.go
@@ -713,15 +713,16 @@ func (gm *golangManager) attemptBinaryRebuild(
 		binaryNeedsStdlib := false
 		if stdlibFixedVersion != "" {
 			binaryGoVersion := strings.TrimPrefix(binaryInfo.GoVersion, "go")
-			if binaryGoVersion == "" {
+			switch {
+			case binaryGoVersion == "":
 				// Synthetic binaries from distroless fallback have no Go version info.
 				// Assume they need stdlib rebuild since we can't prove otherwise.
 				binaryNeedsStdlib = true
 				log.Infof("  Binary %s has unknown Go version (synthetic), assuming stdlib rebuild needed", binaryPath)
-			} else if isValidGoVersion(binaryGoVersion) && isLessThanGoVersion(binaryGoVersion, stdlibFixedVersion) {
+			case isValidGoVersion(binaryGoVersion) && isLessThanGoVersion(binaryGoVersion, stdlibFixedVersion):
 				binaryNeedsStdlib = true
-				log.Debugf("  Binary %s (Go %s) needs stdlib upgrade to >= %s", binaryPath, binaryGoVersion, stdlibFixedVersion)
-			} else {
+				log.Infof("  Binary %s (Go %s) needs stdlib upgrade to >= %s", binaryPath, binaryGoVersion, stdlibFixedVersion)
+			default:
 				log.Debugf("  Binary %s (Go %s) already has stdlib >= %s, no stdlib rebuild needed", binaryPath, binaryGoVersion, stdlibFixedVersion)
 			}
 		}
--- a/pkg/langmgr/golang.go
+++ b/pkg/langmgr/golang.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 
 	"github.com/moby/buildkit/client/llb"
+	gwclient "github.com/moby/buildkit/frontend/gateway/client"
 	"github.com/project-copacetic/copacetic/pkg/buildkit"
 	"github.com/project-copacetic/copacetic/pkg/provenance"
 	"github.com/project-copacetic/copacetic/pkg/types/unversioned"
@@ -168,6 +169,7 @@ func filterGoPackages(langUpdates unversioned.LangUpdatePackages) (unversioned.L
 				continue
 			}
 			goPackages = append(goPackages, pkg)
+			log.Debugf("filterGoPackages: keeping %s@%s → %s (type=%s, path=%s)", pkg.Name, pkg.InstalledVersion, pkg.FixedVersion, pkg.Type, pkg.PkgPath)
 		}
 	}
 	return goPackages, stdlibFixedVersion
@@ -650,6 +652,7 @@ func (gm *golangManager) attemptBinaryRebuild(
 
 	// Build update map from updates (shared across all binaries)
 	updateMap := make(map[string]string)
+	log.Debugf("[updateMap] building from %d updates", len(updates))
 	for _, update := range updates {
 		if update.FixedVersion == "" {
 			log.Debugf("Skipping %s: no fixed version available", update.Name)
@@ -799,7 +802,25 @@ func (gm *golangManager) attemptBinaryRebuild(
 			continue
 		}
 
-		// Success - update state for next iteration
+		// Verify the rebuild actually works by doing an intermediate Solve.
+		// RebuildBinary only constructs the LLB graph — it doesn't execute it.
+		// Without this check, Copa logs success but the build may fail later.
+		verifyDef, verifyErr := newState.Marshal(ctx)
+		if verifyErr != nil {
+			log.Warnf("Failed to marshal rebuild state for %s (skipping): %v", binaryPath, verifyErr)
+			rebuildFailures = append(rebuildFailures, rebuildFailure{binaryPath: binaryPath, reason: fmt.Sprintf("marshal error: %v", verifyErr)})
+			continue
+		}
+		_, solveErr := gm.config.Client.Solve(ctx, gwclient.SolveRequest{
+			Definition: verifyDef.ToPB(),
+		})
+		if solveErr != nil {
+			log.Warnf("Binary rebuild for %s built LLB successfully but execution failed (skipping): %v", binaryPath, solveErr)
+			rebuildFailures = append(rebuildFailures, rebuildFailure{binaryPath: binaryPath, reason: fmt.Sprintf("build execution failed: %v", solveErr)})
+			continue
+		}
+
+		// Verified — update state for next iteration
 		state = &newState
 		totalRebuilt++
 		log.Debugf("Prepared rebuild for binary: %s", binaryPath)
--- a/pkg/langmgr/golang.go
+++ b/pkg/langmgr/golang.go
@@ -545,10 +545,12 @@ func (f rebuildFailure) String() string {
 func collectGoBinaryInfo(langUpdates unversioned.LangUpdatePackages) (paths []string, goVersion string) {
 	seen := make(map[string]bool)
 	for _, u := range langUpdates {
-		if u.Type != utils.GoBinary && u.Type != utils.GoModules {
+		// Only gobinary entries carry binary paths via PkgPath. GoModules entries' PkgPath
+		// points at go.mod/go.sum locations which are not useful for synthetic BinaryInfo.
+		if u.Type != utils.GoBinary {
 			continue
 		}
-		// Extract Go version from stdlib entries (e.g., "v1.26.0" → "1.26.0")
+		// Extract Go version from stdlib entries (e.g., "v1.26.0" -> "1.26.0").
 		if u.Name == "stdlib" && goVersion == "" && u.InstalledVersion != "" {
 			goVersion = strings.TrimPrefix(u.InstalledVersion, "v")
 		}
@@ -683,6 +685,10 @@ func (gm *golangManager) attemptBinaryRebuild(
 	totalAttempted := 0
 	var rebuildFailures []rebuildFailure
 
+	// Image-level metadata reused across every binary in this image. Parse once to avoid
+	// re-unmarshalling the OCI config for every binary.
+	imageSourceLabel := extractOCISourceLabel(gm.config)
+
 	// Process each detected binary
 	for i, binaryInfo := range binaries {
 		binaryPath := binaryInfo.Path
@@ -758,26 +764,34 @@ func (gm *golangManager) attemptBinaryRebuild(
 			}
 		}
 
+		// Resolution paths cloneSourceCode supports: VCS metadata (sourceCommit + sourceRepo),
+		// --go-vcs-url override, OCI source label + image tag, or image-tag heuristic. Only
+		// fail-fast here when none of these can fire; otherwise let cloneSourceCode try and
+		// produce its specific error if every fallback ultimately fails.
+		hasFallback := gm.goVCSURL != "" || imageSourceLabel != "" || gm.imageRef != ""
 		switch {
-		case sourceCommit == "":
-			log.Warnf("  Binary %s has no VCS commit info and no OCI revision label. "+
-				"Cannot rebuild without source.", binaryPath)
-			rebuildFailures = append(rebuildFailures, rebuildFailure{binaryPath: binaryPath, reason: "no source commit available"})
+		case sourceCommit == "" && sourceRepo == "" && !hasFallback:
+			log.Warnf("  Binary %s has no VCS info, no --go-vcs-url, no OCI source label, "+
+				"and no image tag. Cannot rebuild without source.", binaryPath)
+			rebuildFailures = append(rebuildFailures, rebuildFailure{binaryPath: binaryPath, reason: "no source resolution path"})
 			continue
-		case sourceRepo == "":
-			log.Warnf("  Binary %s has commit %s but no source repo could be derived. "+
+		case sourceCommit != "" && sourceRepo == "" && !hasFallback:
+			log.Warnf("  Binary %s has commit %s but no source repo and no fallback. "+
 				"Cannot rebuild without source.", binaryPath, sourceCommit)
 			rebuildFailures = append(rebuildFailures, rebuildFailure{binaryPath: binaryPath, reason: "no source repo"})
 			continue
 		default:
-			log.Debugf("  Source: %s @ %s", sourceRepo, sourceCommit)
+			if sourceCommit != "" && sourceRepo != "" {
+				log.Debugf("  Source: %s @ %s", sourceRepo, sourceCommit)
+			} else {
+				log.Debugf("  Source: deferring resolution to cloneSourceCode (fallback path)")
+			}
 		}
 
 		for module, version := range filteredUpdateMap {
 			log.Debugf("  Will update %s to %s", module, version)
 		}
 
-		// Create rebuild context for this binary
 		rebuildCtx := &provenance.RebuildContext{
 			Strategy:         provenance.RebuildStrategyHeuristic,
 			BuildInfo:        buildInfo,
@@ -785,7 +799,7 @@ func (gm *golangManager) attemptBinaryRebuild(
 			ImageLabels:      gm.config.ImageLabels,
 			ImageRef:         gm.imageRef,
 			GoVCSURL:         gm.goVCSURL,
-			ImageSourceLabel: extractOCISourceLabel(gm.config),
+			ImageSourceLabel: imageSourceLabel,
 		}
 
 		// Attempt to rebuild this binary and merge into current state
--- a/pkg/langmgr/golang.go
+++ b/pkg/langmgr/golang.go
@@ -42,6 +42,7 @@ type golangManager struct {
 	toolchainPatchLevel string
 	goVCSURL            string
 	imageRef            string
+	goBinaryPaths       []string // binary paths from all gobinary updates (including stdlib)
 }
 
 // validateGoPackageName validates a Go module name for safety and correctness.
@@ -184,6 +185,10 @@ func (gm *golangManager) InstallUpdates(
 	// Filter for Go packages only
 	goUpdates, stdlibFixedVersion := filterGoPackages(manifest.LangUpdates)
 	hasStdlib := stdlibFixedVersion != ""
+
+	// Collect binary paths from ALL gobinary updates (including stdlib) for
+	// the synthetic binary fallback on distroless images where detection fails.
+	gm.goBinaryPaths = collectGoBinaryPaths(manifest.LangUpdates)
 
 	// Only act on stdlib vulns if user explicitly opted in via --toolchain-patch-level
 	if hasStdlib && gm.toolchainPatchLevel == "" {
@@ -532,21 +537,30 @@ func (f rebuildFailure) String() string {
 	return fmt.Sprintf("%s: %s", f.binaryPath, f.reason)
 }
 
-// buildSyntheticBinaryInfo constructs BinaryInfo entries from Trivy report data
-// when go version -m detection fails (e.g., distroless/scratch images without a shell).
-// Binary paths come from UpdatePackage.PkgPath, which Trivy sets from the gobinary
-// scan result's Target field.
-func buildSyntheticBinaryInfo(updates unversioned.LangUpdatePackages, goVCSURL string) []*provenance.BinaryInfo {
+// collectGoBinaryPaths extracts unique binary paths from all gobinary updates,
+// including stdlib entries that filterGoPackages strips out. These paths are needed
+// for the synthetic binary fallback when go version -m detection fails.
+func collectGoBinaryPaths(langUpdates unversioned.LangUpdatePackages) []string {
 	seen := make(map[string]bool)
+	var paths []string
+	for _, u := range langUpdates {
+		if (u.Type == utils.GoBinary || u.Type == utils.GoModules) && u.PkgPath != "" && !seen[u.PkgPath] {
+			seen[u.PkgPath] = true
+			paths = append(paths, u.PkgPath)
+		}
+	}
+	return paths
+}
+
+// buildSyntheticBinaryInfo constructs BinaryInfo entries from collected binary paths
+// when go version -m detection fails (e.g., distroless/scratch images without a shell).
+// binaryPaths comes from collectGoBinaryPaths which includes paths from ALL gobinary
+// updates (including stdlib entries stripped by filterGoPackages).
+func buildSyntheticBinaryInfo(binaryPaths []string, goVCSURL string) []*provenance.BinaryInfo {
 	var binaries []*provenance.BinaryInfo
 
-	for _, u := range updates {
-		if u.PkgPath == "" || seen[u.PkgPath] {
-			continue
-		}
-		seen[u.PkgPath] = true
-
-		path := u.PkgPath
+	for _, p := range binaryPaths {
+		path := p
 		if !strings.HasPrefix(path, "/") {
 			path = "/" + path
 		}
@@ -601,7 +615,7 @@ func (gm *golangManager) attemptBinaryRebuild(
 	// `go version -m` cannot run due to missing shell.
 	if (detectErr != nil || len(binaries) == 0) && gm.goVCSURL != "" {
 		log.Info("Binary detection unavailable (distroless/scratch image?), falling back to Trivy report + --go-vcs-url")
-		binaries = buildSyntheticBinaryInfo(updates, gm.goVCSURL)
+		binaries = buildSyntheticBinaryInfo(gm.goBinaryPaths, gm.goVCSURL)
 		if len(binaries) > 0 {
 			log.Infof("Constructed %d synthetic binary entries from Trivy report", len(binaries))
 		}
--- a/pkg/langmgr/golang.go
+++ b/pkg/langmgr/golang.go
@@ -43,6 +43,7 @@ type golangManager struct {
 	goVCSURL            string
 	imageRef            string
 	goBinaryPaths       []string // binary paths from all gobinary updates (including stdlib)
+	goBinaryGoVersion   string   // Go version from stdlib entries (e.g., "1.26.0")
 }
 
 // validateGoPackageName validates a Go module name for safety and correctness.
@@ -188,7 +189,7 @@ func (gm *golangManager) InstallUpdates(
 
 	// Collect binary paths from ALL gobinary updates (including stdlib) for
 	// the synthetic binary fallback on distroless images where detection fails.
-	gm.goBinaryPaths = collectGoBinaryPaths(manifest.LangUpdates)
+	gm.goBinaryPaths, gm.goBinaryGoVersion = collectGoBinaryInfo(manifest.LangUpdates)
 
 	// Only act on stdlib vulns if user explicitly opted in via --toolchain-patch-level
 	if hasStdlib && gm.toolchainPatchLevel == "" {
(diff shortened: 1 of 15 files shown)


pkg/provenance/rebuilder.go
```diff
@@ -438,32 +496,61 @@ func validateCommitHash(commit string) error {
...
+	i := strings.LastIndex(goVCSURL, "@")
+	if i <= 0 || i >= len(goVCSURL)-1 {
+		return "", "", fmt.Errorf("invalid go-vcs-url format, expected repo@ref")
+	}
+	repoURL := goVCSURL[:i]
+	ref := goVCSURL[i+1:]
+	if err := validateRepoURL(repoURL); err != nil {
+		return "", "", err
+	}
+	if err := validateShellSafe(ref, "go-vcs-url ref"); err != nil {
```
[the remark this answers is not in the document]
omercnet: Fixed — switched to `validateShellSafeStrict` for the `--go-vcs-url` ref. Git refs/tags have no reason to contain quotes.

pkg/provenance/rebuilder.go
```diff
@@ -472,26 +559,68 @@ func (r *Rebuilder) cloneSourceCode(buildInfo *BuildInfo) (llb.State, string, er
...
+			return llb.State{}, "", err
+		}
+		ref = commit
+		log.Infof("Cloning source from binary VCS metadata %s @ %s", repoURL, ref)
+	} else {
+		tag := ""
+		if rebuildCtx != nil {
+			tag = extractImageTag(rebuildCtx.ImageRef)
+		}
+		if tag == "" {
```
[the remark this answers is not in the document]
omercnet: Fixed — added `looksLikeSemverTag()` that requires the tag to start with a digit (after optional `v` prefix) and contain at least one dot. Rejects `latest`, `stable`, `alpine`, `sha-abc123`, etc. Added 11 test cases.

pkg/provenance/rebuilder.go
```diff
@@ -472,26 +559,68 @@ func (r *Rebuilder) cloneSourceCode(buildInfo *BuildInfo) (llb.State, string, er
...
+			if repoURL != "" {
+				log.Warnf("No commit/tag specified for %s - skipping clone", repoURL)
+			}
+			return llb.State{}, "", fmt.Errorf("no commit/tag specified for source clone")
+		}
+		if err := validateShellSafe(tag, "image tag"); err != nil {
```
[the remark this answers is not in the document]
omercnet: Fixed — switched to `validateShellSafeStrict` for the image tag as well.

website/docs/app-level-patching.md
```diff
@@ -402,22 +409,42 @@ copa patch \
...
-:::danger Critical Requirement
-Go binaries **must have a VCS commit hash embedded** for Copa to clone the source and rebuild. This is the most important requirement for binary rebuilding to work.
-:::
+Copa resolves where to clone source code using a 3-step fallback chain. Each step is tried in order; the first success wins:
+
+| Priority | Method | When it fires |
+|---|---|---|
+| 1 | `--go-vcs-url` override | User explicitly provides `repo@ref` |
+| 2 | VCS commit from binary | Binary has embedded `vcs.revision` |
+| 3 | Image tag heuristic | No VCS info, but image has a semver-like tag (e.g., `v1.2.3`) |
```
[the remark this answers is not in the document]
omercnet: The docs already said "semver-like tag" — the code now matches via `looksLikeSemverTag()`. No doc change needed.

omercnet:
Note: the only CI failure is `vuln-check`, which is pre-existing on `main` — `govulncheck` reports 2 vulnerabilities in `github.com/docker/docker` affecting `pkg/imageloader/docker.go`, `pkg/utils/mediatype.go`, and `pkg/buildkit/buildkit.go`. None of these files were touched by this PR.

omercnet:
## Found a sequencing bug: `--go-vcs-url` is dead code on distroless images

While testing this PR's fallback resolution on distroless/scratch images, I found that `--go-vcs-url` is never actually read when binary detection fails:

```
attemptBinaryRebuild()
  └─ DetectGoBinaries()         ← go version -m via BuildKit
       └─ FAILS on distroless   ← "no Go binaries detected"
       └─ returns error          ← function exits here

RebuildBinary()                  ← where --go-vcs-url is parsed
  └─ NEVER REACHED
```

The flag is consumed in `cloneSourceCode()` inside `RebuildBinary()`, but that function only runs after detection succeeds. On distroless images (no shell → `go version -m` can't exec), detection fails and the flag is silently ignored.

### Fix

When detection fails and `--go-vcs-url` is provided, construct synthetic `BinaryInfo` from the Trivy report's `PkgPath` field — which already contains the binary path from the gobinary scan. The rebuild machinery (BuildKit LLB diff/merge) already works on distroless; it's only the detection gate that blocks it.

~50 lines in `attemptBinaryRebuild()` + helper. Implementation: [verity-org/copacetic@8a221352](https://github.com/verity-org/copacetic/commit/8a221352)

This should probably be included in this PR since without it, the `--go-vcs-url` flag added here doesn't work on the images that need it most (distroless).

omercnet:
(Updated commit ref after DCO sign-off: [verity-org/copacetic@b4db1913](https://github.com/verity-org/copacetic/commit/b4db1913))

robert-cronin (dismissed):
(removed - posted in error)

pkg/provenance/rebuilder.go
```diff
@@ -384,6 +384,64 @@ func deriveRepoFromModulePath(modulePath string) (repoURL string, subpath string
...
+				subpath = parts[2]
+			}
+			return repoURL, subpath
+		}
+	}
+
+	if strings.HasPrefix(modulePath, "go.opentelemetry.io/") {
+		parts := strings.SplitN(modulePath, "/", 3)
+		if len(parts) >= 2 {
+			repoURL = fmt.Sprintf("https://github.com/open-telemetry/%s", parts[1])
```
robert-cronin, line 421: I think this whole `go.opentelemetry.io/` block needs a lookup table rather than a split-based derivation. The OpenTelemetry project does not follow the `github.com/open-telemetry/<name>` pattern:

- `go.opentelemetry.io/otel` maps to `open-telemetry/otel` here, but the actual repo is `open-telemetry/opentelemetry-go`
- `go.opentelemetry.io/contrib` maps to `open-telemetry/contrib`, actual is `open-telemetry/opentelemetry-go-contrib`
- `go.opentelemetry.io/collector` maps to `open-telemetry/collector`, actual is `open-telemetry/opentelemetry-collector`

A small map from known opentelemetry module prefixes to repo paths would be more accurate. Same probably applies to `go.uber.org/` in practice (e.g., `go.uber.org/zap` does happen to be `uber-go/zap`, but it is convention, not a rule).

pkg/provenance/rebuilder.go
```diff
@@ -472,26 +575,93 @@ func (r *Rebuilder) cloneSourceCode(buildInfo *BuildInfo) (llb.State, string, er
...
+		}
+		ref = commit
+		log.Infof("Cloning source from binary VCS metadata %s @ %s", repoURL, ref)
+	case rebuildCtx != nil && rebuildCtx.ImageSourceLabel != "":
+		// Step 3: OCI label org.opencontainers.image.source provides the source repo.
+		// Combine with image tag as the git ref.
+		labelURL := rebuildCtx.ImageSourceLabel
+		if !strings.HasPrefix(labelURL, "https://github.com/") {
+			return llb.State{}, "", fmt.Errorf("OCI source label URL not on github.com: %s", labelURL)
+		}
```
robert-cronin, line 607: I think this hard-reject is overly restrictive. The OCI source label is something the image author explicitly set, so treating non-github.com hosts as an error (instead of falling through to step 4) surprises users who host on GitLab, Gitea, self-hosted forges, etc.

Option: drop the github.com check here and rely on `validateRepoURL` to allow any trusted host, or log a debug message and fall through to the tag heuristic.

pkg/report/trivy.go
```diff
@@ -432,17 +432,24 @@ func (t *TrivyParser) ParseWithLibraryPatchLevel(file, libraryPatchLevel string)
...
+						key := vuln.PkgName + "\x00" + pkgPath
 						if _, exists := langPackageVulns[key]; !exists {
 							langPackageVulns[key] = []trivyTypes.DetectedVulnerability{}
 							langPackageInfo[key] = unversioned.UpdatePackage{
 								Name:             vuln.PkgName,
 								Type:             string(r.Type),
 								Class:            string(r.Class),
 								InstalledVersion: vuln.InstalledVersion,
-								PkgPath:          vuln.PkgPath,
+							PkgPath:          pkgPath,
```
robert-cronin, line 452: Indentation regressed here - this line lost a tab level relative to the sibling fields (`Name:`, `Type:`, `Class:`, `InstalledVersion:` above). It will probably fail `gofmt` check in CI.

```suggestion
								PkgPath:          pkgPath,
```

website/docs/app-level-patching.md
```diff
@@ -371,6 +371,13 @@ copa patch \
     -r scan.json \
     --pkg-types library \
     --toolchain-patch-level
+
+# Provide source repo for stripped binaries (no embedded VCS info)
+copa patch \
+    -i $IMAGE \
+    -r scan.json \
+    --pkg-types library \
+    --go-vcs-url https://github.com/org/repo@v1.2.3
```
robert-cronin, line 380: `--go-vcs-url` is gated behind `COPA_EXPERIMENTAL=1` in `cmd.go` (it is registered inside the experimental flags branch). Worth calling that out in the docs so users do not hit a confusing "unknown flag" error.

website/docs/app-level-patching.md
```diff
@@ -402,22 +409,42 @@ copa patch \
...
 6. **Main package discovery**: Locates the main package in the cloned source using multiple strategies (explicit path, `cmd/` directory, binary name matching)
 7. **Rebuild and replace**: Rebuilds the binary with updated dependencies and copies it back to its original location in the image
 
-##### Key Requirement: VCS Commit Hash
+##### Source Resolution Chain
 
-:::danger Critical Requirement
-Go binaries **must have a VCS commit hash embedded** for Copa to clone the source and rebuild. This is the most important requirement for binary rebuilding to work.
-:::
+Copa resolves where to clone source code using a 3-step fallback chain. Each step is tried in order; the first success wins:
```
robert-cronin, line 421: Docs say "3-step fallback chain" and the table below lists 3 priorities, but the implementation in `cloneSourceCode` has 4 branches: `--go-vcs-url` override, VCS commit, OCI source label (`org.opencontainers.image.source`), and image tag heuristic. The OCI label step is undocumented. I think the docs should either add it as step 3 (bumping tag heuristic to step 4) or explain why it is not a user-facing step.

robert-cronin (changes requested):
Main concern here is the overlap with PR #1516 (my branch `fix/go-patching-log-levels`). Five files are touched by both PRs (`golang.go`, `rebuilder.go`, `types.go`, `core.go`, `rebuilder_test.go`). Both solve the same core problem (stripped binaries with no VCS metadata) but at different layers. I think we should merge #1516 first and then rebase this on top, since the approaches are complementary.

One integration note: once #1516 lands (and once the duplicate-parsing point you raised there is addressed), the OCI labels will already be extracted once in `buildkit.go`. This PR's Go VCS fallback logic could read from that cached manifest instead of re-parsing the image config, which would avoid introducing a second parse path.

Specific issues inline. Happy to help coordinate the merge order with #1516.

omercnet:
lets get #1516 merged first and get back to this one
it's fully reviewed, can you fix it you want me to pick it up ?

omercnet:
## Reply to PR #1546 (robert-cronin)

Top-level summary comment on the PR (not inline):

---

Thanks for the thorough review. Pushed `badde0b` addressing the inline feedback:

**Fixed in code:**
- `pkg/report/trivy.go:452`: gofmt indentation restored
- `pkg/langmgr/golang.go:673`: if-else chain rewritten as switch (gocritic)
- `pkg/provenance/rebuilder.go:607`: OCI source label step no longer hard-rejects non-github hosts. It now falls through to the tag heuristic when `validateRepoURL` fails, so users on GitLab/Gitea/self-hosted forges are not blocked.
- `pkg/provenance/rebuilder.go:418`: OpenTelemetry block replaced with a lookup table. You were right that `go.opentelemetry.io/otel` does not map to `open-telemetry/otel` - the real repo is `opentelemetry-go`. Added explicit mappings for `otel`, `contrib`, `collector`, and `ebpf-profiler`, with unknown top-level modules falling through to the generic derivation. Tests updated to match.

**Fixed in docs:**
- `website/docs/app-level-patching.md`: table now documents all 4 fallback steps (including the OCI source label step), and the `--go-vcs-url` example notes the `COPA_EXPERIMENTAL=1` requirement.

**Still open / for discussion:**
- On `go.uber.org/`: agreed it is convention, not a rule. I left the split-based derivation for now since the common cases (`zap`, `atomic`, `multierr`, `fx`, `dig`) do follow `uber-go/<name>`. Happy to add a similar lookup table if you prefer the stricter approach - let me know.
- **Merge order with #1516**: makes sense. I am keeping this as a draft and will rebase onto #1516 once it lands. Also happy to cut out the duplicate OCI label parsing in favor of reading from `buildkit.go`'s extracted manifest once #1516 addresses that.

CI should go green on the next run. Pre-existing `vuln-check` failure is the docker/docker issue #1525 addresses.

pkg/provenance/rebuilder.go
omercnet: Fixed in `fd4ff4a` - replaced split-based derivation with an `otelRepos` lookup table covering `otel` -> `opentelemetry-go`, `contrib` -> `opentelemetry-go-contrib`, `collector` -> `opentelemetry-collector`, `ebpf-profiler` -> `opentelemetry-ebpf-profiler`. Unknown top-level modules fall through to generic derivation. See `pkg/provenance/rebuilder.go:554-574`.

pkg/provenance/rebuilder.go
omercnet: Fixed in `76087c2` - the OCI source label case now gates on `validateRepoURL(rebuildCtx.ImageSourceLabel) == nil` as its match condition, so untrusted hosts fall through to the tag heuristic instead of hard-rejecting. Comment at line 752 calls this out explicitly. See `pkg/provenance/rebuilder.go:748-766`.

pkg/report/trivy.go
omercnet: Fixed in `76087c2` - `PkgPath` now aligns with sibling fields at both call sites. See `pkg/report/trivy.go:452` and `:484`.

website/docs/app-level-patching.md
omercnet: Fixed in `76087c2` - added `COPA_EXPERIMENTAL=1` callouts next to the `--go-vcs-url` example (line 376), in the priority table row (line 426), and in the Step 1 prose of the Source Resolution Chain section (line 431).

website/docs/app-level-patching.md
omercnet: Fixed in `76087c2` - docs now describe a 4-step fallback chain: step 3 is the OCI source label (with its trust-host and semver-tag requirements documented), step 4 is the image tag heuristic. Priority table and prose updated to match the implementation. See `website/docs/app-level-patching.md:413-448`.

omercnet:
Rebased past #1516 (now merged) and #1530. Merge conflicts resolved, build/vet/lint/tests all clean locally.

@robert-cronin the inline feedback from your Apr-17 review is addressed and the 5 threads are resolved with per-thread replies pointing at the specific commits. PR body updated to reflect the 4-step chain, OCI source label step, OpenTelemetry lookup table, and new tests. Ready for re-review whenever convenient.

pkg/langmgr/golang.go
```diff
@@ -529,6 +540,60 @@ func (f rebuildFailure) String() string {
 	return fmt.Sprintf("%s: %s", f.binaryPath, f.reason)
 }
 
+// collectGoBinaryInfo extracts unique binary paths and the installed Go version
+// from all gobinary updates, including stdlib entries that filterGoPackages strips.
+func collectGoBinaryInfo(langUpdates unversioned.LangUpdatePackages) (paths []string, goVersion string) {
+	seen := make(map[string]bool)
+	for _, u := range langUpdates {
+		if u.Type != utils.GoBinary && u.Type != utils.GoModules {
```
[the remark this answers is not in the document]
omercnet: Fixed in `ef83dc4` - the type filter in `collectGoBinaryInfo` is narrowed to `utils.GoBinary` only. GoModules entries' `PkgPath` points at `go.mod`/`go.sum` locations and would produce synthetic `BinaryInfo` with paths that are not binaries. Added two regression cases to `TestCollectGoBinaryInfo` covering mixed gobinary+gomod input and all-gomod input.

pkg/langmgr/golang.go
```diff
@@ -693,10 +779,13 @@ func (gm *golangManager) attemptBinaryRebuild(
...
+			Strategy:         provenance.RebuildStrategyHeuristic,
+			BuildInfo:        buildInfo,
+			BinaryInfo:       []*provenance.BinaryInfo{binaryInfo},
+			ImageLabels:      gm.config.ImageLabels,
+			ImageRef:         gm.imageRef,
+			GoVCSURL:         gm.goVCSURL,
+			ImageSourceLabel: extractOCISourceLabel(gm.config),
 		}
 
 		// Attempt to rebuild this binary and merge into current state
```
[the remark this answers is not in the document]
omercnet: Fixed in `ef83dc4` - the gate now considers all four resolution paths (`sourceCommit` + `sourceRepo` from VCS metadata, `--go-vcs-url`, OCI source label, and image tag) and only short-circuits when none can fire. Otherwise it defers to `cloneSourceCode` which produces a specific error if every fallback ultimately fails. The debug log now distinguishes the VCS-resolved case from the deferred-to-fallback case.

pkg/langmgr/golang.go
```diff
@@ -693,10 +779,13 @@ func (gm *golangManager) attemptBinaryRebuild(
...
-			BinaryInfo:  []*provenance.BinaryInfo{binaryInfo},
-			ImageLabels: gm.config.ImageLabels,
+			Strategy:         provenance.RebuildStrategyHeuristic,
+			BuildInfo:        buildInfo,
+			BinaryInfo:       []*provenance.BinaryInfo{binaryInfo},
+			ImageLabels:      gm.config.ImageLabels,
+			ImageRef:         gm.imageRef,
+			GoVCSURL:         gm.goVCSURL,
+			ImageSourceLabel: extractOCISourceLabel(gm.config),
 		}
```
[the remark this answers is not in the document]
omercnet: Fixed in `ef83dc4` - `extractOCISourceLabel(gm.config)` is now called once before the per-binary loop and stashed in a local `imageSourceLabel` variable. The OCI config is parsed once per image instead of once per binary, and the same value is passed into every `RebuildContext.ImageSourceLabel`.

pkg/provenance/rebuilder.go
```diff
@@ -623,11 +807,11 @@ func (r *Rebuilder) cloneSourceCode(buildInfo *BuildInfo) (llb.State, string, er
 	if !strings.HasSuffix(gitRef, ".git") {
 		gitRef += ".git"
 	}
-	if commit != "" {
-		gitRef += "#" + commit
+	if ref != "" {
+		gitRef += "#" + ref
 	}
```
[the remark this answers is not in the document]
omercnet: Fixed in `ef83dc4` - the final `Cloning source from %s @ %s` debug log in `cloneSourceCode` now uses `ref` instead of `commit`. `commit` is empty for the `--go-vcs-url` override, OCI source label, and image tag fallback paths, so the log was misleading. The two case-arm logs above already use `ref`, so this brings the trailing log into line.

pkg/langmgr/golang.go
```diff
@@ -953,3 +1074,27 @@ func (gm *golangManager) upgradePackagesWithTooling(
...
+		} `json:"config"`
+	}
+	if err := json.Unmarshal(config.ConfigData, &imageConfig); err != nil {
+		log.Debugf("Could not parse image config for OCI labels: %v", err)
+		return ""
+	}
+
+	source := imageConfig.Config.Labels["org.opencontainers.image.source"]
+	if source != "" {
+		log.Debugf("Found OCI label org.opencontainers.image.source: %s", source)
```
robert-cronin, line 1097: This duplicates work already done by `pkg/buildkit/buildkit.go:extractLabelsFromConfig`, which populates `config.ImageLabels` for exactly this purpose. Elsewhere in this file we already read OCI annotations via `gm.config.ImageLabels[ociAnnotationRevision]` / `[ociAnnotationSource]`.

Suggest replacing the whole helper with a single lookup at the call site:
```go
imageSourceLabel := ""
if gm.config != nil && gm.config.ImageLabels != nil {
    imageSourceLabel = gm.config.ImageLabels["org.opencontainers.image.source"]
}
```
or add a constant alongside the other `ociAnnotation*` constants and use it. Avoids the second JSON unmarshal and a subtle inconsistency: `extractLabelsFromConfig` uses `json:"Labels"` (capital L per the OCI spec) while this helper uses `json:"labels"` — Go's case-insensitive matching saves us, but they should agree.

pkg/provenance/rebuilder.go
```diff
@@ -574,32 +643,77 @@ func validateCommitHash(commit string) error {
...
+		return false
+	}
+	// Must start with a digit and contain at least one dot (e.g., "1.2")
+	if t[0] < '0' || t[0] > '9' {
+		return false
+	}
+	return strings.Contains(t, ".")
+}
+
+func parseGoVCSURL(goVCSURL string) (string, string, error) {
```
robert-cronin, line 678: `parseGoVCSURL` is only called from `cloneSourceCode`, which means an invalid `--go-vcs-url` (wrong format, untrusted host, unsafe ref) is not surfaced until BuildKit has already been initialized and solving has started. Could you call `parseGoVCSURL` once in `pkg/cmd/cmd.go` right after flags are parsed (when `ua.goVCSURL != ""`) so users get a fast, clear failure for typos? The validation logic itself stays here; the call site just becomes earlier.

pkg/provenance/rebuilder.go
```diff
@@ -574,32 +643,77 @@ func validateCommitHash(commit string) error {
...
+		return ""
+	}
+	return imageRef[lastColon+1:]
+}
+
+// looksLikeSemverTag returns true if the tag looks like a version string
+// (e.g., "v1.2.3", "1.45.0", "v3.9.1-rc1"). This prevents the image tag
+// heuristic from trying non-version tags like "latest", "stable", or "alpine"
+// as git refs.
+func looksLikeSemverTag(tag string) bool {
```
robert-cronin, line 666: Minor: `looksLikeSemverTag` accepts `1.45` (two-component, no patch). For most upstream Go projects the actual git tag would be `v1.45.0`, so this fallback would fire and then fail at the actual git fetch with a less obvious error. Either tighten this to require at least two dots (`major.minor.patch`-ish) or make the log line at the call site explicit that the tag was used verbatim, so users can self-diagnose.

pkg/provenance/rebuilder.go
```diff
@@ -608,6 +722,76 @@ func (r *Rebuilder) cloneSourceCode(buildInfo *BuildInfo) (llb.State, string, er
...
+		// Step 4: Image tag heuristic — derive repo from module path, use image tag as ref.
+		tag := ""
+		if rebuildCtx != nil {
+			tag = extractImageTag(rebuildCtx.ImageRef)
+		}
+		if tag == "" || !looksLikeSemverTag(tag) {
+			if repoURL != "" {
+				log.Warnf("No commit/tag specified for %s - skipping clone", repoURL)
+			}
+			return llb.State{}, "", fmt.Errorf("no commit/tag specified for source clone")
```
robert-cronin, line 777: When `ImageRef` is digest-pinned (e.g. `repo@sha256:...`) — which is common in production — `extractImageTag` correctly returns `""`, and we end up here with the fairly generic `no commit/tag specified for source clone` error. That case will be a frequent footgun for distroless+digest-pinned users. A more targeted hint along the lines of `image is digest-pinned and binary has no VCS info; use --go-vcs-url to specify the source ref` would save a lot of head-scratching.

pkg/langmgr/golang.go
```diff
@@ -529,6 +540,62 @@ func (f rebuildFailure) String() string {
...
+			modulePath = strings.TrimPrefix(goVCSURL[:i], "https://")
+		}
+
+		log.Infof("  Synthetic binary: %s (module: %s)", path, modulePath)
+		binaries = append(binaries, &provenance.BinaryInfo{
+			Path:          path,
+			GoVersion:     goVersion, // from stdlib InstalledVersion or empty
+			ModulePath:    modulePath,
+			Dependencies:  make(map[string]string),
+			BuildSettings: map[string]string{"CGO_ENABLED": "0"},
```
robert-cronin, line 590: `buildSyntheticBinaryInfo` derives `modulePath` from `goVCSURL` by stripping `https://` and chopping at the last `@`. That works, but it loses the ability to handle binaries from *different* modules than the override URL points at — every synthetic entry gets the same module path. If a distroless image legitimately ships multiple Go binaries from different repos, that limitation should at least be a documented caveat (and the user has to run copa once per binary with different `--go-vcs-url` values).

robert-cronin (approved):
Thanks for the thorough rework — confirmed all 5 items from my earlier request-changes review are resolved (otel lookup table, OCI label fallthrough on non-github hosts, gofmt indentation, COPA_EXPERIMENTAL callouts in docs, 4-step fallback chain documented). Approving on that basis.

Five smaller follow-up nits below from a fresh read + running unit tests/lint (all green). None are blocking — happy to see these landed in this PR or a follow-up:

- `extractOCISourceLabel` in `pkg/langmgr/golang.go` re-parses the OCI image config JSON, but `gm.config.ImageLabels` is already populated by `extractLabelsFromConfig` in `pkg/buildkit/buildkit.go` and the same file already reads other OCI annotations via `gm.config.ImageLabels[ociAnnotationRevision]`. Worth using the same pattern to keep one source of truth.
- The OCI-label fallthrough (good fix!) means non-github hosts cleanly drop to step 4, but step 4 still calls `validateRepoURL` which is github.com-only — so GitLab/Gitea users still need `--go-vcs-url`. Worth being explicit about that in the docs.
- `--go-vcs-url` is only validated inside `cloneSourceCode`, mid-build. A bad value surfaces only after BuildKit init. A cheap up-front `parseGoVCSURL` check in `cmd.go` would give a much faster, clearer failure for typos.
- `looksLikeSemverTag` accepts 2-component tags like `1.45`, which usually fail at git fetch with a less obvious error. Could tighten or improve the error.
- Digest-pinned `ImageRef` (common in prod) hits the generic "no commit/tag specified" path. A targeted "image is digest-pinned; use `--go-vcs-url`" hint would save head-scratching.

PR #1547, opened 2026-04-11
feat: add experimental Helm chart-aware patching

### LINKED ISSUE #1494: [REQ] Helm chart-aware patching: `copa patch-chart`

New feature

### Summary

Today, Copa patches individual container images. However, the dominant unit of deployment in Kubernetes is the **Helm chart** — a versioned bundle of manifests, values, and sub-charts that together describe a complete application. This proposal requests a new `copa patch-chart` subcommand that treats Helm charts as first-class citizens, patching all referenced container images and producing an updated chart with corrected image references, all within the same OCI registry.

---

### Motivation

When securing a Kubernetes workload, patching individual images is only half the job. The other half is ensuring that the Helm chart used to deploy those workloads actually references the patched images. Today, users must manually correlate Copa's patched image outputs back to their chart values — a fragile, error-prone, and often undocumented step.

This gap means that even when Copa is run correctly, a subsequent `helm upgrade` using the original chart will re-deploy the unpatched images. Chart-aware patching closes this loop end-to-end.

---

### Proposed CLI Interface

```bash
copa patch-chart \
  --chart oci://registry.example.com/myorg/mychart:1.2.3 \
  --output oci://registry.example.com/myorg/mychart:1.2.3-patched \
  --image-registry registry.example.com/myorg \
  --scanner trivy \
  --ignore-errors
```

**Key flags:**

| Flag | Description |
|---|---|
| `--chart` | OCI reference or local path to the Helm chart |
| `--output` | OCI reference or local path to write the patched chart |
| `--image-registry` | Registry namespace where patched images will be pushed |
| `--tag-suffix` | Suffix appended to patched image tags (default: `-patched`) |
| `--values` | Additional values files to use when rendering the chart for image discovery |
| `--scanner` | Scanner plugin to use (default: `trivy`) |
| `--ignore-errors` | Continue if individual images cannot be patched |
| `--dry-run` | Print discovered images and planned changes without patching |

---

### Proposed Behavior

**Step 1 — Image discovery via `helm template`**

Rather than heuristically parsing `values.yaml`, the command renders the chart using `helm template` and extracts all `image:` fields from the resulting Kubernetes manifests. This approach:
- Correctly handles conditional templates and feature flags
- Resolves sub-chart image references
- Respects any `--values` overrides provided by the user

**Step 2 — Scan and patch each image**

For each discovered image, Copa runs the existing patch pipeline:
1. Scan with the configured scanner (e.g., Trivy)
2. Patch via `copa patch` if vulnerabilities are found
3. Push the patched image to the target registry with a deterministic new tag or digest

Images that are already clean (no fixable CVEs) are skipped; their references remain unchanged in the output chart.

**Step 3 — Generate patched chart**

Copa produces an updated Helm chart where all patchable image references in `values.yaml` (and sub-chart values) are rewritten to point to the patched images. The output is one of:

- **Option A (recommended):** A new chart version pushed to the OCI registry, with `values.yaml` updated in-place and `Chart.yaml` version bumped (e.g., appending `+copa.YYYYMMDD` to the version).
- **Option B:** A `values-copa-patched.yaml` overlay file that users apply with `helm install -f values-copa-patched.yaml`, leaving the original chart untouched.

Both options should be supported. Option B is useful when the user does not control the chart's registry.

**Step 4 — Emit a patch report**

Copa outputs a machine-readable report (JSON) summarizing:
- Which images were discovered
- Which were patched and what CVEs were fixed
- Which were skipped (clean or unpatchable)
- The old→new image reference mapping

This report can optionally be attached to the output chart as an OCI referrer (using the OCI referrers API), enabling downstream attestation and audit workflows.

---

### Image Reference Schema Discovery

Helm charts have no enforced schema for how images are declared. Common patterns include:

```yaml
# Pattern A: split repository/tag
image:
  repository: myregistry.io/myapp
  tag: "1.2.3"

# Pattern B: full image URI
image: myregistry.io/myapp:1.2.3

# Pattern C: digest-pinned
image:
  repository: myregistry.io/myapp
  digest: sha256:abc123...
```

The `helm template`-based discovery approach handles all of these naturally, since it reads from rendered Pod specs rather than raw values. Copa then traces each rendered image URI back to its source location in `values.yaml` to rewrite it.

For cases where automatic tracing is ambiguous (e.g., images constructed dynamically in templates), Copa should:
1. Log a warning identifying the unresolvable reference
2. Continue patching the remaining images
3. Document the unresolved reference in the patch report

---

### OCI-Native Bundle (Stretch Goal)

As a follow-on, Copa could leverage the OCI referrers API to produce a fully self-contained bundle:

```
oci://registry.example.com/myorg/mychart:1.2.3-patched
  └── [helm chart artifact]            ← updated chart with patched image refs
  └── [referrer] patched image A       ← co-located patched image
  └── [referrer] patched image B       ← co-located patched image
  └── [referrer] copa-patch-report     ← attestation / audit trail
```

This would make the patched chart and all its runtime dependencies addressable and distributable as a single OCI artifact — particularly valuable for air-gapped environments.

---

### Why This Belongs in Copa Core

This feature is a natural extension of Copa's existing value proposition:

> *Reduce the complexity of patching [...] to running a single tool.*

Today, "a single tool" still leaves users with a gap between patched images and the charts that deploy them. `copa patch-chart` completes that picture. The scanning, patching, and OCI push logic already exist in Copa — the new work is chart rendering, image reference rewriting, and chart artifact publishing.

A reference implementation of this pattern exists in [Verity](https://github.com/verity-project/verity), an automated container security patching system for Helm charts built on top of Copa and Trivy. The intent is to validate the approach there and contribute the chart-native primitives upstream to Copa.

---

## Are you willing to implement it?

Yes, I already implemented my own concept in my Copa fork used in https://github.com/verity-org/verity but would like to upstream it back here with input from the copa maintainers 

omercnet:
## Summary

Resolves #1494

Adds experimental Helm chart-aware patching to Copa. Given a Helm chart, Copa discovers its rendered container images, patches them, then generates and pushes a self-contained wrapper chart with patched image references.

## What changed

- Added `pkg/helm` for OCI/HTTP chart download, rendering, image discovery, value-path resolution, wrapper construction, packaging, and OCI push.
- Added single-chart mode to `copa patch` behind `COPA_EXPERIMENTAL=1`.
- Extended bulk `PatchConfig` with `charts`, `chartTarget`, and `overrides`.
- Preserved patched images in their original repositories; `--chart-registry` controls only the wrapper chart destination.
- Embedded the original chart in the wrapper artifact so it installs without a dependency update.
- Added deterministic support for scalar `image`, `name/tag`, `repository/tag`, and `registry/repository/tag` value schemas.
- Included hook workloads, init containers, sidecars, ephemeral containers, and Kubernetes List resources in image discovery.
- Rejected digest-pinned images instead of silently changing them to `:latest`.
- Failed chart generation on missing or ambiguous value paths instead of publishing partial overrides.
- Preserved existing patched-image mappings when clean images are skipped on reruns.
- Propagated cancellation through Helm render, discovery, package, and push boundaries.

## Usage

```bash
export COPA_EXPERIMENTAL=1
copa patch \
  --chart reloader \
  --chart-version 1.2.1 \
  --chart-repo oci://ghcr.io/stakater/charts \
  --chart-registry oci://ghcr.io/myorg/charts \
  --push
```

Chart mode requires `--push`. `--image`, `--config`, and `--chart` are mutually exclusive.

## Verification

- `go test -race ./pkg/helm ./pkg/bulk ./pkg/cmd`
- `make test`
- `make build`
- `golangci-lint run ./pkg/helm/... ./pkg/bulk/... ./pkg/cmd/...`
- Docusaurus production build
- Live OCI chart discovery and value-path resolution with Reloader 1.2.1


pkg/cmd/cmd.go
```diff
@@ -124,9 +150,18 @@ copa patch --config copa-bulk-config.yaml --push (Bulk Image Patching)`,
 
 				return bulk.PatchFromConfig(context.Background(), ua.configFile, opts)
```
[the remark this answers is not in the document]
omercnet: Fixed — both `PatchFromConfig` and `PatchChart` now use the signal-cancelled `ctx` from the command handler.

pkg/bulk/engine.go
```diff
@@ -77,6 +95,193 @@ func buildTargetRepository(sourceImage, targetRegistry string) (string, error) {
...
+	var resolutions []chartResolution
+	var errs *multierror.Error
+
+	for _, chartSpec := range charts {
+		log.Infof("Downloading Helm chart '%s' v%s from %s...", chartSpec.Name, chartSpec.Version, chartSpec.Repository)
+		ch, err := helm.DownloadChart(chartSpec.Name, chartSpec.Version, chartSpec.Repository)
+		if err != nil {
+			errs = multierror.Append(errs, fmt.Errorf("chart '%s': %w", chartSpec.Name, err))
+			continue
+		}
```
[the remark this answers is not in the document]
omercnet: Fixed — `resolveChartImagesWithCharts` now checks `ctx.Err()` before processing each chart.

pkg/helm/render.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+
+	// If there's a slash after the last colon, it's a port (e.g., "registry:5000/image")
+	// and there's no explicit tag.
+	if strings.Contains(imageRef[lastColon:], "/") {
+		return imageRef, "latest"
+	}
+
+	return imageRef[:lastColon], imageRef[lastColon+1:]
+}
```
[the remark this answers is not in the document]
omercnet: Fixed — `parseImageRef` now strips `@sha256:...` digest suffixes before tag parsing. Regression test added.

pkg/helm/chart.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+	if len(entries) == 0 {
+		return nil, fmt.Errorf("no chart archive found after pulling '%s'", name)
+	}
+
+	chartPath := tmpDir + "/" + entries[0].Name()
+	ch, err := loader.Load(chartPath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to load chart archive '%s': %w", chartPath, err)
+	}
```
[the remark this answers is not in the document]
omercnet: Fixed — `DownloadChart` now filters for `.tgz` entries instead of blindly using `entries[0]`. Test added.

pkg/bulk/chart.go
```diff
@@ -0,0 +1,176 @@
...
+	res := resolutions[0]
+	log.Infof("Discovered %d image(s) in chart '%s'", len(res.Images), chartSpec.Name)
+
+	// Step 2: Compute target image refs for each discovered image.
+	// Strip the oci:// prefix from the chart registry to get a valid container image registry.
+	// ChartRegistry keeps oci:// for Helm push operations; image refs must not have it.
+	imageRegistry := strings.TrimPrefix(opts.ChartRegistry, "oci://")
+	target := TargetSpec{
+		Registry: imageRegistry,
+	}
```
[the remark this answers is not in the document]
omercnet: By design — in single-chart mode, `--chart-registry` is used to derive the image target registry because the chart is the unit of deployment. Images and wrapper chart go to the same OCI namespace.

pkg/bulk/engine.go
```diff
@@ -77,6 +95,193 @@ func buildTargetRepository(sourceImage, targetRegistry string) (string, error) {
...
+	merged := make([]ImageSpec, len(config.Images))
+	copy(merged, config.Images)
+
+	for _, chartImg := range chartImages {
+		if _, exists := explicitRefs[chartImg.Image]; exists {
+			log.Debugf("Skipping chart-discovered image '%s': overridden by explicit image spec", chartImg.Image)
+			continue
+		}
+		merged = append(merged, chartImg)
+		explicitRefs[chartImg.Image] = struct{}{}
```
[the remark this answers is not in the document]
omercnet: This is intentional — explicit image specs take precedence over chart-discovered images for the same repository. This allows users to override chart defaults with specific tag strategies.

pkg/bulk/config.go
```diff
@@ -44,6 +68,57 @@ type TagStrategy struct {
...
+}
+
+// validateOverrides validates all OverrideSpec entries in the config.
+func validateOverrides(overrides map[string]OverrideSpec) error {
+	for key, o := range overrides {
+		if o.From == "" {
+			return fmt.Errorf("overrides[%q]: from is required", key)
+		}
+		if o.To == "" {
+			return fmt.Errorf("overrides[%q]: to is required", key)
```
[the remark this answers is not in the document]
omercnet: The `from`/`to` fields are always required — `valuePath` is supplementary (provides explicit values.yaml path). An override with only `valuePath` and no `from`/`to` has no tag substitution to apply.

pkg/helm/render.go
```diff
@@ -0,0 +1,137 @@
...
+				result = append(result, img)
+			}
+		}
+	}
+
+	if result == nil {
+		result = []ChartImage{}
+	}
+	return result, nil
+}
```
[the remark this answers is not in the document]
omercnet: Acknowledged — the error return is there for forward compatibility (e.g., strict mode that fails on malformed manifests). Currently all parse failures are logged and skipped, which is correct for chart rendering where partial results are preferable.

pkg/helm/patchedchart.go
```diff
@@ -0,0 +1,287 @@
...
+					RepositoryPath: candidate.path,
+					TagPath:        candidate.tagPath,
+				})
+				matched[img.Repository] = true
+				break
+			}
+		}
+		if !matched[img.Repository] {
+			log.Warnf("helm: could not auto-detect values.yaml path for image %q — use overrides.valuePath to specify manually", img.Repository)
+		}
```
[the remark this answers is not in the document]
omercnet: Map iteration order does not affect correctness here — findRepositoryPaths returns the first matching candidate per image, and Helm charts almost never have duplicate repository keys pointing to the same image.

pkg/helm/patchedchart.go
```diff
@@ -0,0 +1,287 @@
...
+
+	// Build the subchart override map
+	subchartValues := make(map[string]interface{})
+
+	for _, vp := range valuePaths {
+		patched, found := findPatchedMapping(vp.ImageRepo, patchedLookup)
+		if !found {
+			log.Debugf("helm: no patched mapping found for image %q, skipping value override", vp.ImageRepo)
+			continue
+		}
```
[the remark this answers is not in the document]
omercnet: Each chart image has a unique repository in practice — same repo with different tags would be an unusual chart structure. Current behavior (last-write-wins) is acceptable for this edge case.

omercnet:
CI notes:

- **lint**: Fixed — extracted repeated test YAML to a constant (`goconst`).
- **vuln-check**: Pre-existing on `main` — `govulncheck` reports 2 vulnerabilities in `github.com/docker/docker` (`pkg/imageloader/docker.go`, `pkg/utils/mediatype.go`, `pkg/buildkit/buildkit.go`). Not caused by this PR.
- **dependency-review**: `github.com/go-errors/errors@1.4.2` (transitive dep from Helm SDK) has OpenSSF Scorecard of 2.6, below repo threshold of 3.0. This is a transitive dependency of `helm.sh/helm/v3`, not something we can control.
- **snyk (license + security)**: Likely flagging the new Helm SDK dependency tree. Managed by Copa maintainers.

omercnet:
@sozercan @ashnamehrotra @robert-cronin friendly ping - this PR has been open ~16 days. Current blocker is `dependency-review`: `helm.sh/helm/v3` transitively pulls `github.com/go-errors/errors` (OpenSSF Scorecard 2.6 < repo threshold 3).

Three paths I see:
1. Allowlist the transitive dep in repo dep-review settings (low effort, slight policy weakening for one package)
2. Restructure `pkg/helm` to avoid `helm/v3/pkg/action` (it's the path that pulls in `k8s.io/cli-runtime` -> `kustomize` -> `go-errors`); larger refactor
3. Park the PR until upstream Helm drops the dep

Happy to take any direction. Otherwise can leave parked.

omercnet:
Pushed `f76170b1` to bump `helm.sh/helm/v3` from 3.20.0 → 3.20.2, which addresses [GHSA-hr2v-4r36-88hr](https://github.com/advisories/GHSA-hr2v-4r36-88hr) (the moderate-severity vuln the `dependency-review` check was flagging). `go mod tidy` also pulled `k8s.io/{client-go,component-base,kubectl}` from 0.35.0 → 0.35.1.

Verified locally: `go build ./...` clean, `go test ./pkg/helm/... -count=1` passes.

The two Snyk checks (`license/snyk`, `security/snyk`) were failing before this bump and continue to fail; I don't have access to the Copa Snyk org dashboard to see what specifically they're flagging. Could a maintainer link me to the Snyk findings, or confirm whether they're known issues?

omercnet:
Quick follow-up - since the helm 3.20.2 bump landed in `f76170b1` and the `dependency-review` go-errors blocker is still the open question: @sozercan @ashnamehrotra @robert-cronin which of the three paths from the Apr-28 comment do you prefer? Happy to draft the change, or convert this PR to draft until guidance lands.

pkg/helm/patchedchart.go
```diff
@@ -0,0 +1,287 @@
...
+			Version:     origVersion + "-patched.1",
+			Description: fmt.Sprintf("Copa-patched version of %s %s", origName, origVersion),
+			Type:        "application",
+			Annotations: map[string]string{
+				"copa.sh/source-chart":      origName,
+				"copa.sh/source-version":    origVersion,
+				"copa.sh/source-repository": chartSpec.Repository,
+				"copa.sh/patched-at":        time.Now().UTC().Format(time.RFC3339),
+			},
+			Dependencies: []*helmchart.Dependency{
```
robert-cronin: Could we embed the original chart before packaging? This only adds dependency metadata, so the wrapper may install with a missing dependency error.

pkg/cmd/cmd.go
```diff
@@ -110,25 +115,55 @@ copa patch --config copa-bulk-config.yaml --push (Bulk Image Patching)`,
...
+				return bulk.PatchFromConfig(ctx, ua.configFile, opts)
 			}
-			if ua.appImage == "" {
-				return errors.New("--image is required when not using --config")
+
+			if hasChart {
+				if ua.chartVersion == "" || ua.chartRepo == "" {
+					return errors.New("--chart requires --chart-version and --chart-repo")
+				}
+				if ua.chartRegistry == "" {
```
robert-cronin: Could chart mode require `--push`? Otherwise the chart is published while its patched images only exist locally.

pkg/bulk/chart.go
```diff
@@ -0,0 +1,174 @@
...
+	if len(resolutions) == 0 {
+		return fmt.Errorf("chart '%s' could not be downloaded or rendered", chartSpec.Name)
+	}
+
+	res := resolutions[0]
+	log.Infof("Discovered %d image(s) in chart '%s'", len(res.Images), chartSpec.Name)
+
+	// Step 2: Compute target image refs for each discovered image.
+	// Strip the oci:// prefix from the chart registry to get a valid container image registry.
```
robert-cronin: Should patched images have a separate destination? `--chart-registry` is documented for charts, but it also becomes the image registry here.

pkg/helm/render.go
```diff
@@ -0,0 +1,141 @@
...
+
+// parseImageRef splits an image reference string into repository and tag components.
+// If no tag is present, "latest" is returned as the tag.
+// Returns empty strings for an empty input.
+func parseImageRef(imageRef string) (repo, tag string) {
+	if imageRef == "" {
+		return "", ""
+	}
+
+	if digestIndex := strings.Index(imageRef, "@sha256:"); digestIndex != -1 {
```
robert-cronin: Could digest-only references be preserved or rejected? This turns `repo@sha256:…` into `repo:latest`, which may patch a different image.

pkg/helm/chart.go
```diff
@@ -0,0 +1,187 @@
...
+	// charts with a high kubeVersion constraint (e.g. >=1.25.0) don't fail.
+	// ClientOnly mode defaults to v1.20.0 via chartutil.DefaultCapabilities.
+	install.KubeVersion = kubeVersionFromBuildDeps()
+
+	release, err := install.Run(ch, map[string]interface{}{})
+	if err != nil {
+		return "", fmt.Errorf("failed to render chart '%s': %w", ch.Metadata.Name, err)
+	}
+
+	return release.Manifest, nil
```
robert-cronin: Could we include `release.Hooks` too? `release.Manifest` omits hook Jobs, so their images won’t be discovered.

pkg/bulk/engine.go
```diff
@@ -77,6 +95,199 @@ func buildTargetRepository(sourceImage, targetRegistry string) (string, error) {
...
+		img := config.Images[i]
+		explicitRefs[img.Image] = struct{}{}
+	}
+
+	merged := make([]ImageSpec, len(config.Images))
+	copy(merged, config.Images)
+
+	for i := range chartImages {
+		chartImg := chartImages[i]
+		if _, exists := explicitRefs[chartImg.Image]; exists {
```
robert-cronin: Could this dedupe by repository and tag, or merge the tag lists? A chart using the same repository at two tags currently schedules only one.

pkg/bulk/engine.go
```diff
@@ -245,6 +487,13 @@ func PatchFromConfig(ctx context.Context, configPath string, opts *types.Options
 					log.Errorf("Failed to patch %s: %v", imageWithTag, err)
 				} else {
 					jobResult.Status = "Patched"
+					imageMappings = append(imageMappings, ChartImageMapping{
```
robert-cronin: Should this still add a mapping for the existing `ResolvedTag`? On reruns, a clean patched image is skipped and then omitted from the chart.

pkg/bulk/engine.go
```diff
@@ -77,6 +95,199 @@ func buildTargetRepository(sourceImage, targetRegistry string) (string, error) {
...
+
+	for i := range resolutions {
+		res := &resolutions[i]
+		chartMappings := filterMappingsForChart(res, mappings)
+		if len(chartMappings) == 0 {
+			log.Infof("No successfully patched images for chart '%s', skipping chart generation", res.Spec.Name)
+			continue
+		}
+
+		valuePaths := helm.ResolveImageValuePaths(res.Chart.Values, res.Images, explicitPaths)
```
robert-cronin: Could missing paths for any patched mapping return an error? A partial result may publish a chart that leaves some patched images unused.

pkg/bulk/config.go
```diff
@@ -44,6 +68,57 @@ type TagStrategy struct {
...
+
+// isValidChartRepository checks whether a repository URL has a supported scheme.
+func isValidChartRepository(repo string) bool {
+	return strings.HasPrefix(repo, "oci://") || strings.HasPrefix(repo, "https://")
+}
+
+// validateOverrides validates all OverrideSpec entries in the config.
+func validateOverrides(overrides map[string]OverrideSpec) error {
+	for key, o := range overrides {
+		if o.From == "" {
```
robert-cronin: Could `valuePath` be allowed without `from` and `to`? The manual-path escape hatch currently requires an unrelated tag rewrite.

pkg/helm/patchedchart.go
robert-cronin: Could we embed the original chart before packaging? I think this only adds dependency metadata, so the wrapper may fail to install.

pkg/cmd/cmd.go
robert-cronin, line 169: I think chart mode probably needs to require `--push`; otherwise it can publish a chart while the patched images only exist locally.

pkg/helm/render.go
robert-cronin: Could digest-only references be preserved or rejected here? Stripping the digest seems to make them fall back to `:latest`, which could patch a different image.

pkg/helm/chart.go
robert-cronin: `release.Manifest` seems to exclude hook Jobs, so I think this may miss their images. It may be worth including `release.Hooks` too.

pkg/bulk/engine.go
robert-cronin, line 541: Should we also add a mapping for an existing clean `ResolvedTag`? On reruns, I think it can otherwise be skipped and omitted from the chart.

pkg/bulk/engine.go
robert-cronin: It looks like this only handles the case where no paths are found. If some images lack a value path, we may publish partial mappings; I think that should probably be an error.

robert-cronin (commented):
Sorry for the delay here. I left a few comments, but overall I think the approach looks good. Thanks for the contribution!

pkg/helm/patchedchart.go
omercnet: Fixed in ac3b5d9b. The wrapper now embeds the original chart with SetDependencies, and a regression test saves and reloads the archive to verify the dependency is packaged.

pkg/cmd/cmd.go
omercnet: Fixed in ac3b5d9b. Single-chart mode now requires --push, and chartTarget bulk configs also reject publication unless patched images are pushed.

pkg/helm/render.go
omercnet: Fixed in ac3b5d9b. Digest-pinned references are rejected with an actionable error instead of being converted to :latest; tagged and digest-only cases have regression coverage.

pkg/helm/chart.go
omercnet: Fixed in ac3b5d9b. RenderChart now appends release.Hooks manifests, with a hook Job regression test.

pkg/bulk/engine.go
omercnet: Fixed in ac3b5d9b. Skipped clean images now add a ChartImageMapping using ResolvedTag so rerun wrappers retain the existing patched image.

pkg/bulk/engine.go
omercnet: Fixed in ac3b5d9b. Value-path resolution now returns an error for every missing or ambiguous mapped image, so partial wrapper charts are not published.

pkg/helm/patchedchart.go
```diff
@@ -0,0 +1,332 @@
...
+		}
+
+		repository := patched.PatchedRepo
+		if valuePath.RegistryPath != "" {
+			registry, remainder := splitRegistry(patched.PatchedRepo)
+			setNestedValue(subchartValues, valuePath.RegistryPath, registry)
+			repository = remainder
+		}
+		setNestedValue(subchartValues, valuePath.RepositoryPath, repository)
+		if valuePath.TagPath != "" {
```
robert-cronin: Could we fail here when `TagPath` is empty? For charts with `image.repository` but no default `image.tag`—for example, when the template falls back to `.Chart.AppVersion`—the wrapper can still render the original tag instead of the patched one.

pkg/bulk/engine.go
```diff
@@ -279,6 +553,13 @@ func PatchFromConfig(ctx context.Context, configPath string, opts *types.Options
 	// Print a summary of all patch jobs.
 	printSummary(results)
 
+	if config.ChartTarget != nil && len(chartResolutions) > 0 {
```
robert-cronin, line 582: I think we should avoid publishing a wrapper when one of that chart’s patch jobs failed. `imageMappings` only contains successful jobs, so without `--ignore-errors` this can push a partially patched wrapper and then return the patch error afterward.

pkg/helm/patchedchart.go
```diff
@@ -0,0 +1,332 @@
...
+	overrideValues, err := buildOverrideValues(origName, mappings, valuePaths)
+	if err != nil {
+		return nil, err
+	}
+
+	patchedChart := &helmchart.Chart{
+		Metadata: &helmchart.Metadata{
+			APIVersion:  helmchart.APIVersionV2,
+			Name:        origName + "-patched",
+			Version:     origVersion + "-patched.1",
```
robert-cronin: Should the wrapper version advance when the patched image tags do? This always republishes `<source>-patched.1`; an immutable registry will reject a later wrapper, while a mutable one makes the same chart version point to different content.

pkg/helm/patchedchart.go
```diff
@@ -0,0 +1,332 @@
...
+			Description: fmt.Sprintf("Copa-patched version of %s %s", origName, origVersion),
+			Type:        "application",
+			Annotations: map[string]string{
+				"copa.sh/source-chart":      origName,
+				"copa.sh/source-version":    origVersion,
+				"copa.sh/source-repository": chartSpec.Repository,
```
robert-cronin: Could we sanitize the source repository before storing it here? A private URL containing userinfo or a signed query would be copied into the annotation and dependency metadata, then published inside the wrapper chart.

pkg/helm/patchedchart.go
omercnet: Fixed in 0eb84d28. Wrapper generation now fails when a non-scalar image value has no writable tag path, with regression coverage for charts that derive tags elsewhere.

pkg/bulk/engine.go
omercnet: Fixed in 0eb84d28. Bulk patching now tracks failed image sources and refuses to publish every affected chart wrapper, with a mixed success/failure regression test.

pkg/helm/patchedchart.go
omercnet: Fixed in 0eb84d28. Wrapper versions now include a deterministic hash of the canonical image mappings, so changed patched tags produce a new immutable chart version.

pkg/helm/patchedchart.go
omercnet: Fixed in 0eb84d28. Published source repository metadata and dependency fields now strip URL userinfo, query strings, and fragments; regression coverage verifies credentials are not retained.

go.mod
```diff
@@ -39,63 +39,111 @@ require (
...
 	github.com/vito/progrock v0.10.1
 	golang.org/x/exp v0.0.0-20260603202125-055de637280b
-	golang.org/x/mod v0.36.0
-	golang.org/x/sync v0.21.0
+	golang.org/x/mod v0.37.0
+	golang.org/x/sync v0.22.0
 	golang.org/x/term v0.44.0
 	google.golang.org/grpc v1.82.1
 	gopkg.in/yaml.v3 v3.0.1
+	helm.sh/helm/v3 v3.20.2
```
robert-cronin: Could we move this to `helm.sh/helm/v4 v4.1.4` instead of keeping the v3 dependency and waiving the security checks? I tested that migration against this head: after adapting the v4 chart paths/APIs (`NewPull`, the three-argument `Configuration.Init`, `DryRunClient`, and release/hook accessors) and running `go mod tidy`, containerd v1 drops out, `govulncheck ./...` reports zero called vulnerabilities, and the Helm/bulk/cmd tests pass. v4.1.4 is the newest v4 release compatible with the Go 1.25 baseline (v4.2 requires Go 1.26). Could we make that change and use both Snyk checks as acceptance criteria too?

robert-cronin: Could we move this to `helm.sh/helm/v4 v4.1.4` and pin the Snyk-fixed transitive versions? I tested this against the current head: after adapting the v4 APIs, requiring `github.com/distribution/distribution/v3@v3.1.1` and `filippo.io/edwards25519@v1.1.1`, and running `go mod tidy`, Go selects `otlploghttp@v0.19.0`; the Helm/bulk/cmd tests pass and `govulncheck ./...` reports zero called vulnerabilities. v4.1.4 is the newest v4 release compatible with Go 1.25, and these versions meet all 12 Snyk security fixed-in thresholds. Could we use both Snyk checks and `govulncheck` being green as the acceptance criteria here?

robert-cronin: Could we avoid adding Helm as a Go SDK here? The Snyk results show this dependency adds 12 security findings plus two new MPL-2.0 dependencies: `github.com/mitchellh/cli` through `sql-migrate@v1.8.1`, and `github.com/go-sql-driver/mysql` through `gorp@v3.1.0`. Helm v4 addresses the security findings, but it is still monolithic and retains those license dependencies. I think the clean path may be an injectable, context-aware Helm command runner for `pull`, `template --include-crds`, `package`, and `push`, with local tar/YAML handling for `Chart.yaml` and `values.yaml`. Could we use argv directly rather than a shell, return an actionable error when Helm is unavailable, document the prerequisite, then remove the direct `helm.sh/helm` dependency and run `go mod tidy`? I would use the Helm/bulk/cmd tests, `govulncheck`, and both Snyk checks being green as the acceptance criteria.

omercnet: Fixed in 7d5aeab8. Copa now invokes an injectable, context-aware Helm CLI runner without a shell; it locally reads chart metadata and values, embeds the pulled archive, and uses Helm for pull/template/package/push. The Helm SDK and its SQL/license dependency chain are removed, `helm` is documented as the chart-patching prerequisite, and unit tests assert CLI argv plus cancellation and missing-binary behavior.

robert-cronin (dismissed):
lgtm, thank you! looks like there are some conflicts before we can merge

robert-cronin (approved):

PR #1557, opened 2026-04-20
fix(langmgr): suppress NU1605 in generated patch.csproj for .NET

### LINKED ISSUE #1556: [BUG] Test .NET patching CI job fails with NU1605 package downgrade error on multi-package fixes
### Version of copa

main (observed on commit 928892c1 / PR #1532 merge, also present on every PR for weeks)

### Expected Behavior

The `Test .NET patching` CI job (`test-dotnet`, runs `./test/e2e/dotnet`) should succeed when patching a .NET image that has multiple vulnerable NuGet packages where one of the fixed packages transitively depends on a higher version of another fixed package.

Specifically, `TestDotNetSDKImagePatching` on `docker.io/ashnam/dotnet-sdk-vuln:v1` should apply all Trivy-reported fixes without NuGet restore failures.

### Actual Behavior

The job fails consistently on both `TestDotNetSDKImagePatching` and `TestDotNetRuntimeImagePatching`. Recent runs on main (push to `928892c1`, 2026-04-20) and every open PR show the same failure signature. `@robert-cronin` also flagged this as a pre-existing unrelated bug on #1532.

The generated `patch.csproj` contains direct `PackageReference` entries for each Trivy-reported fixed version. One of those references is for a higher-level package whose transitive dependency requires a newer minor version of another package that is also in the list. NuGet then produces an `NU1605` "package downgrade" error and, because the SDK image treats warnings as errors for this rule, `dotnet restore` fails and the whole patch fails.

Condensed log excerpt from `Test .NET patching` job `72126614085`:

```
time="...": level=debug msg="updates to apply: ... [
  {Newtonsoft.Json 12.0.1 13.0.1 CVE-2024-21907}
  {NuGet.Packaging 6.11.1-rc.2 7.3.1 GHSA-g4vj-cjjj-v7hg}
  {NuGet.Protocol 6.11.1-rc.2 7.3.1 GHSA-g4vj-cjjj-v7hg}
  {System.Security.Cryptography.Xml 8.0.0 10.0.6 CVE-2026-26171}
  ...
]"

    <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
    <PackageReference Include="NuGet.Packaging" Version="7.3.1" />
    ...

/patch/patch.csproj : error NU1605: Warning As Error: Detected package downgrade: Newtonsoft.Json from 13.0.3 to 13.0.1. Reference the package directly from the project to select a different version.
/patch/patch.csproj : error NU1605:  patch -> NuGet.Packaging 7.3.1 -> Newtonsoft.Json (>= 13.0.3)
/patch/patch.csproj : error NU1605:  patch -> Newtonsoft.Json (>= 13.0.1)

FAIL test/e2e/dotnet.TestDotNetSDKImagePatching
```

Job log: https://github.com/project-copacetic/copacetic/actions/runs/24666775526/job/72126614085

### Steps To Reproduce

1. Check out `main` at commit `928892c1` or later.
2. Run the `.NET` e2e suite locally, pointing `copa` at a BuildKit endpoint:
   ```bash
   export COPA_EXPERIMENTAL=1
   make build
   gotestsum --format testname -- ./test/e2e/dotnet --addr=docker:// --copa="$(pwd)/copa" -timeout 0
   ```
   Or watch the `Test .NET patching` job on any recent PR.
3. Observe `TestDotNetSDKImagePatching` fail with `error NU1605` from `dotnet restore` inside the patch SDK container.

### Root Cause

In `pkg/langmgr/dotnet.go`, `patchRuntimeImage` builds `patch.csproj` by appending a `<PackageReference Include="X" Version="Y" />` for each fixed version returned by Trivy. When the set of fixed versions includes a package whose transitive dependency graph requires a *newer* version of another package that is also in the set, NuGet restore emits `NU1605` and, because that rule is promoted to error, restore fails.

Example from the failing run:
- `Newtonsoft.Json` is pinned to `13.0.1` (Trivy-reported fix for `CVE-2024-21907`).
- `NuGet.Packaging 7.3.1` (another Trivy-reported fix) has a transitive dependency `Newtonsoft.Json >= 13.0.3`.
- NuGet sees this as a downgrade from `13.0.3` to `13.0.1` and fails.

### Proposed Fix

Let NuGet's natural "highest-wins" resolution handle the conflict, which is safe here because:
- The stated goal of the patch is to reach a non-vulnerable version.
- A higher version is by definition at least as secure as the Trivy-suggested fix.
- The existing `GetUniqueLatestUpdates` already picks the newest fix per package, so there is no risk of silently downgrading past a known fix.

Two minimal options for the generated `patch.csproj`:

1. Add `<NoWarn>NU1605</NoWarn>` to the `<PropertyGroup>`. NuGet then raises `13.0.1` to `13.0.3` automatically and restore succeeds.
2. Add `<TreatWarningsAsErrors>false</TreatWarningsAsErrors>` plus explicit `<WarningsNotAsErrors>NU1605</WarningsNotAsErrors>` for belt-and-suspenders.

Either change is a handful of lines in `dotnet.go` and should unblock the `Test .NET patching` CI job immediately.

I can open a PR with option 1 plus a regression test that asserts the generated `patch.csproj` suppresses `NU1605` and that `TestDotNetSDKImagePatching` passes against the existing test images.

omercnet:
Fixes #1556

## What

The `.NET` patch pipeline in `pkg/langmgr/dotnet.go` generates a minimal `patch.csproj` with one `<PackageReference>` per Trivy-reported fixed version, then runs `dotnet restore + publish` inside an SDK container to extract the patched DLLs. When two of the reported fixes have transitive relationships in NuGet's dependency graph, the direct pin on one package conflicts with the transitive minimum required by another. NuGet raises `NU1605` "package downgrade", the rule is treated as error by default, and `dotnet restore` fails. This breaks the entire patch.

Observed failure from `Test .NET patching` on `main` (commit `928892c1`):

```
<PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
<PackageReference Include="NuGet.Packaging" Version="7.3.1" />

/patch/patch.csproj : error NU1605: Warning As Error: Detected package downgrade: Newtonsoft.Json from 13.0.3 to 13.0.1.
  patch -> NuGet.Packaging 7.3.1 -> Newtonsoft.Json (>= 13.0.3)
  patch -> Newtonsoft.Json (>= 13.0.1)
```

## Fix

Add `<NoWarn>NU1605</NoWarn>` to the generated csproj so NuGet's natural highest-wins resolution picks the transitive minimum automatically. This is strictly safer than the Trivy-suggested fix:

- The resolved version is always `>=` the reported fixed version, never below a known good version.
- `GetUniqueLatestUpdates` already selects the newest fix per package, so there is no risk of silently downgrading past a known fix.

Also extracted the csproj rendering into a `buildPatchCsproj` helper so the logic can be unit tested without spinning up BuildKit.

## Test plan

- `go test ./pkg/langmgr/...` passes locally, including two new unit tests for `buildPatchCsproj` asserting that `<NoWarn>NU1605</NoWarn>` lives inside `<PropertyGroup>` and that the ItemGroup is well-formed even when empty.
- `golangci-lint run --new-from-rev=upstream/main` reports 0 issues.
- Once CI runs against `docker.io/ashnam/dotnet-sdk-vuln:v1` and `docker.io/ashnam/dotnet-runtime-vuln:v2`, the `Test .NET patching` job that has been failing on every recent PR should pass again.

Signed-off-by: Omer &lt;omer@descope.com&gt;

### DIFF
--- a/pkg/langmgr/dotnet.go
+++ b/pkg/langmgr/dotnet.go
@@ -79,6 +79,31 @@ func escapeXMLAttribute(s string) string {
 	return replacer.Replace(s)
 }
 
+// buildPatchCsproj renders the minimal csproj used to download patched
+// NuGet packages inside the SDK container.
+//
+// NU1605 is suppressed because Trivy reports a minimum fixed version per
+// package, but different fixed packages can have transitive dependencies on
+// each other with higher minimum versions (e.g. NuGet.Packaging 7.3.1 depends
+// transitively on Newtonsoft.Json >= 13.0.3 while Trivy also pins
+// Newtonsoft.Json to 13.0.1 as its direct fix). Without NoWarn NuGet flags
+// that as a "downgrade" and dotnet restore fails with NU1605.
+// Letting NuGet's highest-wins resolution pick the transitive minimum is
+// strictly safer than the Trivy-suggested fix - the resolved version is
+// always >= the Trivy fix, never below a known good version.
+func buildPatchCsproj(targetFramework, packageRefs string) string {
+	return fmt.Sprintf(`<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>%s</TargetFramework>
+    <OutputType>Library</OutputType>
+    <NoWarn>NU1605</NoWarn>
+  </PropertyGroup>
+  <ItemGroup>
+%s  </ItemGroup>
+</Project>
+`, targetFramework, packageRefs)
+}
+
 // isValidDotnetVersion checks if a version string is a valid NuGet version.
 // NuGet supports Major.Minor.Patch[.Revision][-prerelease][+buildmetadata].
 func isValidDotnetVersion(v string) bool {
@@ -359,18 +384,10 @@ func (dnm *dotnetManager) patchRuntimeImage(
 	}
 	targetFramework := fmt.Sprintf("net%s", tfmVersion)
 
-	// Create complete project file in one command
+	patchCsproj := buildPatchCsproj(targetFramework, packageRefs.String())
 	createProjectCmd := fmt.Sprintf(`sh -c 'mkdir -p /patch && cat > /patch/patch.csproj << "EOF"
-<Project Sdk="Microsoft.NET.Sdk">
-  <PropertyGroup>
-    <TargetFramework>%s</TargetFramework>
-    <OutputType>Library</OutputType>
-  </PropertyGroup>
-  <ItemGroup>
-%s  </ItemGroup>
-</Project>
-EOF
-cat /patch/patch.csproj'`, targetFramework, packageRefs.String())
+%sEOF
+cat /patch/patch.csproj'`, patchCsproj)
 
 	projectCreated := sdkState.Run(
 		llb.Shlex(createProjectCmd),
--- a/pkg/langmgr/dotnet.go
+++ b/pkg/langmgr/dotnet.go
@@ -595,13 +595,13 @@ echo "Updating %s: %s -> (resolve from generated deps.json, Trivy pin %s)"
 # Find the actual resolved package key in the generated deps.json. NU1605 is
 # suppressed during restore so NuGet may have resolved to a version newer than
 # the Trivy pin if a transitive dep required it.
-RESOLVED_KEY=$(jq -r --arg name "%s" '.targets[$GEN_TARGET_KEY] | keys[] | select(startswith($name + "/"))' /output/patch.deps.json 2>/dev/null | head -1)
+RESOLVED_KEY=$(jq -r --arg name "%s" --arg targetKey "$GEN_TARGET_KEY" '.targets[$targetKey] | keys[] | select(startswith($name + "/"))' /output/patch.deps.json 2>/dev/null | head -1)
 RESOLVED_LIB_KEY=$(jq -r --arg name "%s" '.libraries | keys[] | select(startswith($name + "/"))' /output/patch.deps.json 2>/dev/null | head -1)
 
 if [ -n "$RESOLVED_KEY" ] && [ "$RESOLVED_KEY" != "null" ] && [ -n "$RESOLVED_LIB_KEY" ] && [ "$RESOLVED_LIB_KEY" != "null" ]; then
 	RESOLVED_VERSION="${RESOLVED_KEY#%s/}"
 	echo "  Resolved version: $RESOLVED_VERSION"
-	NEW_TARGET=$(jq -r --arg key "$RESOLVED_KEY" '.targets[$GEN_TARGET_KEY][$key]' /output/patch.deps.json 2>/dev/null)
+	NEW_TARGET=$(jq -r --arg key "$RESOLVED_KEY" --arg targetKey "$GEN_TARGET_KEY" '.targets[$targetKey][$key]' /output/patch.deps.json 2>/dev/null)
 	NEW_LIBRARY=$(jq -r --arg key "$RESOLVED_LIB_KEY" '.libraries[$key]' /output/patch.deps.json 2>/dev/null)
 	echo "  Extracted package metadata from generated deps.json"
 
--- a/pkg/langmgr/dotnet.go
+++ b/pkg/langmgr/dotnet.go
@@ -578,51 +578,75 @@ echo "Generated target framework: $GEN_TARGET_KEY"
 	for _, update := range safeUpdates {
 		packageName := update.Name
 		oldVersion := update.InstalledVersion
-		newVersion := update.FixedVersion
+		trivyPin := update.FixedVersion
 
+		// Resolve the actual package key from the generated deps.json rather than
+		// assuming it is "Name/TrivyPin". NU1605 is suppressed in the generated
+		// patch.csproj (see buildPatchCsproj), so NuGet's highest-wins resolution
+		// can legitimately produce a higher version than Trivy requested when a
+		// transitive dependency requires it. Looking up by "Name/TrivyPin"
+		// in that case returns empty and the deps.json merge is skipped, leaving
+		// stale hashPath/sha512 referencing the old version while the DLL on
+		// disk has been upgraded - an inconsistent manifest.
 		fmt.Fprintf(&script, `
-# Update %s from %s to %s
-echo "Updating %s: %s -> %s"
+# Update %s (Trivy requested >= %s, actual resolved version looked up from generated deps.json)
+echo "Updating %s: %s -> (resolve from generated deps.json, Trivy pin %s)"
 
-# Extract just the package entries from generated deps.json
-NEW_TARGET=$(jq -r ".targets[\"$GEN_TARGET_KEY\"][\"%s/%s\"] // empty" /output/patch.deps.json 2>/dev/null)
-NEW_LIBRARY=$(jq -r '.libraries["%s/%s"] // empty' /output/patch.deps.json 2>/dev/null)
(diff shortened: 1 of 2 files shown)


pkg/langmgr/dotnet.go
```diff
@@ -79,6 +79,31 @@ func escapeXMLAttribute(s string) string {
...
+// always >= the Trivy fix, never below a known good version.
+func buildPatchCsproj(targetFramework, packageRefs string) string {
+	return fmt.Sprintf(`<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>%s</TargetFramework>
+    <OutputType>Library</OutputType>
+    <NoWarn>NU1605</NoWarn>
+  </PropertyGroup>
+  <ItemGroup>
+%s  </ItemGroup>
```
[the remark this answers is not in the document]
omercnet: Thanks for reviewing. I verified this against the actual output and the current template is correct:

**Empty `packageRefs`:**
```xml
  <ItemGroup>
  </ItemGroup>
```
Two leading spaces, not three. `TestBuildPatchCsproj_EmptyPackageRefs` passes locally (and in the new CI run for this PR).

**With refs:**
```xml
  <ItemGroup>
    <PackageReference Include="A" Version="1.0" />
    <PackageReference Include="B" Version="2.0" />
  </ItemGroup>
```
Also clean - the caller always appends each ref with a trailing `\n`, so the `%s  </ItemGroup>` pattern lands `</ItemGroup>` on its own line.

The suggested change (`%s\n  </ItemGroup>`) would actually introduce a blank line before `</ItemGroup>` whenever `packageRefs` is non-empty, since refs already end with `\n`. Keeping the current template.

pkg/langmgr/dotnet.go
```diff
@@ -79,6 +79,31 @@ func escapeXMLAttribute(s string) string {
...
+// package, but different fixed packages can have transitive dependencies on
+// each other with higher minimum versions (e.g. NuGet.Packaging 7.3.1 depends
+// transitively on Newtonsoft.Json >= 13.0.3 while Trivy also pins
+// Newtonsoft.Json to 13.0.1 as its direct fix). Without NoWarn NuGet flags
+// that as a "downgrade" and dotnet restore fails with NU1605.
+// Letting NuGet's highest-wins resolution pick the transitive minimum is
+// strictly safer than the Trivy-suggested fix - the resolved version is
+// always >= the Trivy fix, never below a known good version.
+func buildPatchCsproj(targetFramework, packageRefs string) string {
+	return fmt.Sprintf(`<Project Sdk="Microsoft.NET.Sdk">
```
robert-cronin, line 95: Heads-up (not blocking this PR): with NU1605 now suppressed, the resolved package version in `/output/patch.deps.json` can legitimately differ from `update.FixedVersion` (Trivy's pin) when a transitive dep forces a higher minimum. The `%s/%s` lookup on lines 569-570 uses the Trivy pin, so it'll return empty for those packages and this guard will skip the deps.json merge. The DLL still gets copied (since publish emits the resolved version), but deps.json will reference the old version with stale hashPath/sha512. Consider resolving the version by scanning keys under `.targets[$k]` that start with `${Name}/`.

robert-cronin (commented):
Mostly happy to approve this, NU1605 suppression is the right call and the extraction of `buildPatchCsproj` + unit tests is a nice quality-of-life improvement.

One real concern worth raising before merging: with NU1605 silenced, NuGet's highest-wins resolution will now pick a *different* version than Trivy's pin whenever there's a transitive conflict (which is precisely the case this PR targets, Newtonsoft.Json 13.0.1 → 13.0.3 in the example). That's strictly safer for the restore itself, but I think it quietly breaks the downstream `buildUpdateDepsJsonScript` step. That script looks up `${Name}/${Trivy_FixedVersion}` in the generated `patch.deps.json` (see `pkg/langmgr/dotnet.go:569-570`), and when the resolved version differs, the lookup returns empty → the `if` guard at :572 fails → the deps.json merge is skipped for that package. Net effect: the patched image gets the newer DLL physically (good) but `deps.json` still references the old version with the original `hashPath`/`sha512`. At runtime the loader will typically still load the DLL by path, but the manifest is factually inconsistent with the file on disk.

The existing .NET e2e test image (`ashnam/dotnet-sdk-vuln:v1`) probably doesn't exercise the transitive-conflict case (only Newtonsoft.Json vulnerable, no NuGet.Packaging), which is why CI goes green despite the gap.

Not a blocker for me if you want to merge the restore fix now and handle the deps.json side in a follow-up, the current behaviour is still better than `dotnet restore` failing outright. But worth tracking explicitly. Fix direction would be to resolve the version from `/output/patch.deps.json` (e.g. `.targets[$k] | keys[] | select(startswith("${Name}/"))`) rather than assuming `Name/TrivyPin`.

Everything else checks out, build passes, unit tests pass locally, and the XML output looks correct including the heredoc-safety probe I ran.

pkg/langmgr/dotnet.go
omercnet: Great catch - this was a real latent inconsistency between the DLL on disk and the deps.json manifest. Fixed in e0dbf41.

The script now resolves the actual package key from `/output/patch.deps.json` via `jq .targets[$GEN_TARGET_KEY] | keys[] | select(startswith($name + "/"))` and strips the `$Name/` prefix to get the resolved version. That resolved key is then used for the target lookup, the library lookup, and the dependency-version rewrite across all targets. When NuGet bumps `Newtonsoft.Json` from the Trivy pin `13.0.1` to the transitive minimum `13.0.3`, both the DLL and the manifest entry end up on `13.0.3` consistently.

Added three regression tests in `TestBuildUpdateDepsJsonScript_*` asserting that the script no longer hardcodes `Name/TrivyPin` and does use the `startswith($name + "/")` scan.

Also rebased onto current `main` to pick up #1537.

robert-cronin (approved):
Thanks Omer. Pushed a 2-line fix as `8cfe314f` directly to this branch (maintainer-edits was on):

The `RESOLVED_KEY` and `NEW_TARGET` jq programs referenced `$GEN_TARGET_KEY` as a jq variable but only had it set as a shell var — jq errored `$GEN_TARGET_KEY is not defined`, `2>/dev/null` hid it, every package fell into the `ERROR: Could not find <name> in generated deps.json` branch. Net result: NU1605 was fixed, but the deps.json was never actually updated, so Trivy still reported the original CVE after patching.

Fix passes `--arg targetKey "$GEN_TARGET_KEY"` and references `$targetKey` inside the jq program, mirroring the `$ORIG_TARGET_KEY` pattern already used a few lines down.

Verified locally against the real fixtures:
- `dotnet-runtime-vuln:v2`: 2 → 1 vulns, Newtonsoft CVE-2024-21907 cleared
- `dotnet-sdk-vuln:v1`: 9 → 8 vulns, Newtonsoft CVE-2024-21907 cleared
- Unit tests still pass

Approving on the assumption CI confirms the same. Great debugging on the NU1605 root cause — the fallback resolution logic is the right call.

One non-blocker for a follow-up: the existing unit tests `assert.Contains` substrings of the generated script but don't actually pipe it through `jq` against a fixture `patch.deps.json`. A small executable test would have caught the undefined-jq-var regression. Happy to file or contribute.

omercnet:
This PR is approved by @robert-cronin and all checks except `Test .NET patching` pass - that failure is the pre-existing regression tracked in #1556, which this PR fixes.

Could a maintainer take a look and merge when ready? The `Test .NET patching` CI failure is currently blocking PRs #1525, #1546, #1547, #1552, and #1558 as well, so this unblocks review-ready work across the queue.

PR #1574, opened 2026-04-30
fix(frontend): bump frontend.Dockerfile Go base to 1.25.9-alpine3.23

omercnet:
Refs #1570.

## Problem

`frontend.Dockerfile` pins `golang:1.25.5-alpine3.23` as its builder base. `go.mod` requires `go >= 1.25.9`, so `go build` inside the builder fails with:

```
go: go.mod requires go >= 1.25.9 (running go 1.25.5; GOTOOLCHAIN=local)
```

This blocks the `copacetic-frontend` image build on every tag push. Discovered by @robert-cronin while cutting `v0.14.0-rc.1`.

## Fix

Bump the base image to `golang:1.25.9-alpine3.23`. One-line change in `frontend.Dockerfile`.

## Verification

```sh
docker buildx build --platform linux/amd64 -f frontend.Dockerfile -t copa-frontend-test:1570 .
# build succeeds; final image size 27.8MB (consistent with prior builds)
```

## Scope

Addresses the build-blocker portion of #1570 only. The other acceptance items in that issue (drift-prevention guard, release.md post-tag verification doc, `.goreleaser.yml` `prerelease: auto`) are intentionally left for follow-up PRs to keep this one minimal and reviewable.

PR #1572, opened 2026-04-29
ci: add trusted fork test workflow with ok-to-test label gate

ashnamehrotra:
- New [Maintainer] Trusted Fork Tests workflow (trusted-fork-tests.yml)
- Runs on pull_request_target labeled event with ok-to-test label
- Verifies actor has write/maintain/admin permission
- Creates temporary upstream branch and dispatches Build + Gateway Tests
- Deletes temp branch after dispatch
- Documentation in CONTRIBUTING.md for maintainers

Describe the changes in this pull request using active verbs such as _Add_, _Remove_, _Replace_ ...

Closes #<_issue_ID_>

PR #1581, opened 2026-04-30
ci: harden trusted-fork dispatch against fork-mutated workflow YAML and force-push races

robert-cronin:
Follow-up to #1572 / #1573.

While testing the new fork-PR CI flow on a real fork PR (#1574), two latent issues with the temp-branch approach surfaced:

1. **Workflow YAML was being read from the fork's temp branch.** `gh workflow run --ref ci/trusted/pr-N` evaluates the workflow YAML on the temp branch, which is just the fork's HEAD. A malicious PR could rewrite `build.yml`/`test.yml` itself and have those changes execute on self-hosted runners with secrets.
2. **GITHUB_TOKEN can't push fork-modified workflow files.** `refusing to allow a GitHub App to create or update workflow .github/workflows/trusted-fork-tests.yml without workflows permission` — the orchestrator failed entirely whenever the PR touched any `.github/workflows/` file.

Both stem from the same root cause: pushing fork code as a real branch and dispatching against that branch.

## What this PR does

Switches to a Prow-style pattern: orchestrator dispatches `build.yml`/`test.yml` against the **PR's base branch** (so the workflow YAML always comes from a trusted ref) and passes the fork PR's head SHA + number as inputs. The dispatched workflows then check out fork code by `refs/pull/<n>/head` as data, with `persist-credentials: false`, and re-validate the SHA before any heavy job runs.

### `trusted-fork-tests.yml`
- Drops temp-branch creation, the 10s sleep, the temp-branch deletion, and `contents: write` permission.
- Re-queries PR head SHA + base ref + state via the API at orchestration time (narrows the force-push race window vs. trusting the label-event payload).
- Dispatches `build.yml`/`test.yml` with `--ref <PR base ref>` and `pr_sha`/`pr_number` inputs.
- Idempotent label removal (single step, doesn't fail if already absent).
- Concurrency group per PR with `cancel-in-progress: false`.

### `build.yml` and `test.yml`
- Add `workflow_dispatch.inputs` (`pr_sha`, `pr_number`).
- Add `run-name` showing PR number + SHA on trusted-fork dispatches.
- New `validate-dispatch` job (GitHub-hosted, ~30s) that:
  - validates input shape (`pr_sha` is 40-char hex, `pr_number` is digits),
  - rejects mismatched-pair inputs,
  - re-queries PR via API and asserts `state == open`, `head.sha == pr_sha`, `base.ref == github.ref_name` — closes the dispatch-to-runner-pickup window where a force-push could otherwise substitute unreviewed code.
- All self-hosted jobs gain `needs: validate-dispatch` (transitively via `build`).
- Every `actions/checkout` switches to a single-step ternary: `ref: ${{ inputs.pr_number && format('refs/pull/{0}/head', inputs.pr_number) || '' }}`, `persist-credentials: false`, `fetch-depth: 1`. Empty `ref` falls back to `github.ref`/`github.sha` for normal push/PR runs.
- Every `actions/setup-go` gains `cache: ${{ inputs.pr_number == '' }}` to prevent fork code from poisoning the GHA Go module/build cache used by mainline runs.
- Codecov upload step in `unit-test` is gated on `inputs.pr_number == ''` so `CODECOV_TOKEN` is never exposed to fork code.
- Drops unused `packages: write` on the `build` job (no GHCR publish in this workflow).
- Concurrency group keyed on `inputs.pr_number` with `cancel-in-progress` only for trusted-fork runs.

### `CONTRIBUTING.md`
Adds a 'Trust model' subsection documenting both protected behaviors and explicitly accepted risks (self-hosted runner persistence, matrix-fanout DoS, manual `workflow_dispatch` UI bypass for write users).

## Verification

- `actionlint` clean on all three workflow files (only the unrelated pre-existing self-hosted runner label warning remains).
- YAML parse-clean for all three.
- Job graph preserved: `unit-test` and `build` now `needs: validate-dispatch`; downstream test jobs unchanged (`needs: build` already transitively depends on validate-dispatch).
- Will need a manual trusted-fork dispatch on a fork PR to confirm end-to-end.

## Why not a composite checkout action?

Considered and rejected: a local composite action can't run before the repository is checked out, so it can't be the wrapper for the first checkout step. The single-step ternary pattern accomplishes the same thing without bootstrapping issues.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,6 +21,17 @@ on:
       - "docs/**"
       - "demo/**"
   workflow_dispatch:
+    inputs:
+      pr_sha:
+        description: "Fork PR head SHA. Set by trusted-fork-tests orchestrator."
+        required: false
+        default: ""
+        type: string
+      pr_number:
+        description: "Fork PR number. Set by trusted-fork-tests orchestrator."
+        required: false
+        default: ""
+        type: string
 
 env:
   TRIVY_VERSION: 0.69.3
@@ -43,6 +54,8 @@ jobs:
           egress-policy: audit
       - name: Check out code
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          ref: ${{ inputs.pr_number && format('refs/pull/{0}/head', inputs.pr_number) || '' }}
       - uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.4.0
         with:
           go-version: "1.25.9"
@@ -74,6 +87,8 @@ jobs:
         with:
           egress-policy: audit
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          ref: ${{ inputs.pr_number && format('refs/pull/{0}/head', inputs.pr_number) || '' }}
       - uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.4.0
         with:
           go-version: "1.25.9"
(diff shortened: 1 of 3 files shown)


robert-cronin:
Closing in favor of #1582 — after digging in, the entire `trusted-fork-tests.yml` flow from #1572 duplicates GitHub's built-in **"Approval for running fork pull request workflows from contributors"** setting (Settings → Actions → General). One repo toggle + one click per push replaces ~150 lines of orchestration YAML, a bespoke label, and the permission-check job. #1582 deletes the lot.

PR #1587, opened 2026-05-04
fix: bind media type detection to patch context

sozercan:
### Motivation
- Media type detection used `context.Background()` and remote lookups without a caller context, allowing Docker daemon or registry calls to hang indefinitely and bypass the patch timeout.
- The patch flow called media type resolution before the BuildKit flow, so an unresponsive endpoint could cause a denial-of-service while the patch goroutine remained blocked.

### Description
- Introduces `GetMediaTypeWithContext(ctx, imageRef, runtime)` and keeps `GetMediaType(...)` as a backward-compatible wrapper that calls the new helper with `context.Background()`.
- Threads the caller `ctx` into local and remote resolution: `localMediaType` and `podmanMediaType` now accept a `context.Context` and use it for `cli.ImageInspect` and `exec.CommandContext`, and `remoteMediaType` passes the context to `remote.Get` via `remote.WithContext(ctx)`.
- Updates the patch decision path to call `shouldExportAsOCI(ctx, ...)`, which uses `GetMediaTypeWithContext` so media type detection is bound to the patch timeout/cancellation.
- Adjusts unit tests in `pkg/utils/mediatype_test.go` to use the new context-aware signatures.

### Testing
- Ran `go test ./pkg/utils -run 'MediaType'` and the media-type related unit tests passed.
- Ran `go test ./pkg/utils ./pkg/patch`; some tests failed in this environment due to missing/blocked Docker/registry dependencies rather than the media-type change.
- Attempted `go test` for a patch test (`TestDoesNotExist`) under a timeout; it timed out in this CI-like environment, indicating external integration dependencies rather than the fix itself.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69d7e6c29acc832aaec6bba1f29114a9)


omercnet (commented):
Context wiring looks good. Please add a cancellation regression test for GetMediaTypeWithContext or remoteMediaType so timeout behavior is covered.

robert-cronin (approved):

ashnamehrotra (approved):

PR #1594, opened 2026-05-06
feat(langmgr): scaffold Java/JVM language manager foundation

omercnet:
Establishes the Java/JVM language-manager foundation without implementing JAR replacement yet.

## Changes

- recognizes Trivy `jar`, `pom`, `gradle`, and `sbt` language results
- preserves package type, versions, vulnerability IDs, and target paths in `LangUpdates`
- keeps identical Maven coordinates in different pom/Gradle/sbt manifests distinct by falling back to the Trivy result target when `PkgPath` is empty
- registers one Java manager across all four JVM result types
- reports all Java updates as failed while patching remains unimplemented
- returns an error unless `--ignore-errors` is set, preventing creation of an unchanged patched image by default
- updates Go and the frontend builder fallback to 1.26.6, plus `golang.org/x/crypto` to v0.56.0 and its required `x/*` updates, to resolve current vulnerability findings
- verifies the immutable Chisel tooling image against its own pinned Go 1.25.13 build metadata, decoupled from the repository toolchain version

## Scope

This is foundation only. Maven artifact download, verification, and in-image JAR replacement remain outside this PR.

## Verification

- `gofumpt` on changed Go files
- `go test ./pkg/langmgr/... ./pkg/report/...`
- `make test` with Go 1.26.6
- `go build ./...` with Go 1.26.6
- `golangci-lint run ./...` with Go 1.26.6 and golangci-lint 2.10.1
- `govulncheck ./...`
- `docker buildx build --load -f frontend.Dockerfile .`
- `zizmor .github/workflows/build.yml` produced no findings beyond the 48 already present on `origin/main`
- pinned Chisel metadata verification passed for every platform; local runtime smoke tests could not execute foreign architectures without QEMU


pkg/langmgr/java.go
```diff
@@ -0,0 +1,75 @@
...
+		}
+	}
+	return out
+}
+
+// InstallUpdates is the LangManager entry point. The foundation scaffold logs
+// each affected coordinate, returns the unchanged state, and reports the Java
+// updates as failed packages so the caller can decide whether to hard-fail or
+// continue under --ignore-errors.
+func (jm *javaManager) InstallUpdates(
```
ashnamehrotra, line 53: agree with the copilot comment regarding ignore errors. Also, we should exit non-zero / make sure no -patched image is created when no updates were applied

pkg/report/trivy_test.go
```diff
@@ -1180,3 +1180,52 @@ func writeTempReport(t *testing.T, report *trivyTypes.Report) string {
...
+					{
+						Class: utils.LangPackages,
+						Type:  ftypes.LangType(tc.langType),
+						Vulnerabilities: []trivyTypes.DetectedVulnerability{
+							{
+								VulnerabilityID:  "CVE-2099-9999",
+								PkgName:          tc.pkgName,
+								InstalledVersion: "1.0.0",
+								FixedVersion:     "1.0.1",
+								PkgPath:          "/usr/share/app/lib/" + strings.ReplaceAll(tc.pkgName, ":", "-") + "-1.0.0.jar",
```
ashnamehrotra: where are we getting this path pattern from?

ashnamehrotra:
@omercnet thanks for putting this up! I had a few comments and also agree with the comments put up by Copilot

pkg/report/trivy.go
```diff
@@ -438,7 +438,8 @@ func (t *TrivyParser) ParseWithLibraryPatchLevel(file, libraryPatchLevel string)
 		// Process Language packages
 		if r.Class == utils.LangPackages {
 			// Check if this is a Python, Node.js, or Go related target
-			if r.Type == utils.PythonPackages || r.Type == utils.NodePackages || r.Type == utils.GoModules || r.Type == utils.GoBinary {
+			if r.Type == utils.PythonPackages || r.Type == utils.NodePackages || r.Type == utils.GoModules || r.Type == utils.GoBinary ||
+				r.Type == utils.JavaJar || r.Type == utils.JavaPom || r.Type == utils.JavaGradle || r.Type == utils.JavaSbt {
```
robert-cronin, line 442: Could pom/gradle/sbt fall back to r.Target when PkgPath is empty so the same coordinate in different manifests is not collapsed here?

pkg/report/trivy.go
```diff
@@ -438,7 +438,8 @@ func (t *TrivyParser) ParseWithLibraryPatchLevel(file, libraryPatchLevel string)
 		// Process Language packages
 		if r.Class == utils.LangPackages {
 			// Check if this is a Python, Node.js, or Go related target
```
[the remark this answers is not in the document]
omercnet: Fixed in 1dcde7be: the comment now describes all language package targets supported by Copa.

pkg/langmgr/java.go
```diff
@@ -0,0 +1,75 @@
...
+// reports every Java update as a failed package, so users hitting Java vulns
+// get a clear log message and the orchestrator routes them through
+// --ignore-errors handling rather than silently skipping.
+type javaManager struct {
+	config        *buildkit.Config
+	workingFolder string
+}
+
+// isJavaUpdate returns true when the package type is one of Trivy's Java
+// language types. Used by both the manager and the report parser.
```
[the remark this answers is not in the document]
omercnet: Fixed in 1dcde7be: the helper comment now accurately limits its use to the language manager.

pkg/langmgr/java.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+
+	log.Warnf("Java/JVM library patching is not yet implemented. %d update(s) skipped.", len(javaUpdates))
+	failed := make([]string, 0, len(javaUpdates))
+	for _, u := range javaUpdates {
+		log.Debugf("  Java skipped: %s (installed=%s, fixed=%s, type=%s, path=%s)",
+			u.Name, u.InstalledVersion, u.FixedVersion, u.Type, u.PkgPath)
+		failed = append(failed, u.Name)
+	}
+	return currentState, failed, nil
```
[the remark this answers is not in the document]
omercnet: Fixed in 1dcde7be with regression coverage: unimplemented Java updates now return an error unless --ignore-errors is set, while still reporting every failed coordinate.

pkg/langmgr/java.go
omercnet: Fixed in 1dcde7be: when errors are not ignored the Java scaffold now exits through a non-nil error, preventing an unchanged patched image from being produced. The ignore-errors path remains explicit and tested.

pkg/report/trivy_test.go
omercnet: Updated in 1dcde7be: the test now uses Trivy-style targets for pom/Gradle/sbt and a nested JAR path for jar instead of synthesizing one path pattern for every type.

pkg/report/trivy.go
omercnet: Fixed in 1dcde7be with a regression test: pom, Gradle, and sbt now fall back to Result.Target when PkgPath is empty, so identical coordinates in separate manifests remain distinct.

robert-cronin:
This looks good to me, but I’ll leave the final approval to @ashnamehrotra for an additional review pass in case I missed anything.

PR #1597, opened 2026-05-07
fix(frontend): restore /tmp on scratch image, switch to distroless, build real Dockerfile in e2e

sozercan:
## Summary

Three changes that together fix and prevent the recurrence of a regression in the published copa BuildKit frontend (v0.14.0 / `:latest`), where every patch run fails with:

```
failed to extract report from context:
  failed to create temp dir for report file:
    stat /tmp: no such file or directory
```

### 1. `pkg/frontend/build.go` — restore `/tmp` workaround as a helper

After #1507's security-hardening sweep switched `extractReportFile` / `extractReportDirectory` from `os.MkdirTemp(\"/\", ...)` to `os.MkdirTemp(\"\", ...)` and removed the explicit `os.MkdirAll(\"/tmp\", 0o1777)` that previously paired with it, the v0.14.0 frontend image (built `FROM scratch`) fails on every report extraction. `os.MkdirTemp(\"\")` delegates to `os.TempDir()` (`/tmp` on Linux), which doesn't exist in a scratch rootfs.

Add a small `ensureTempDir()` helper that `MkdirAll`s `os.TempDir()` (mode 0o1777) and call it from both extraction sites before `os.MkdirTemp(\"\", ...)`. Preserves #1507's intent (don't write under `/`) while restoring the runtime invariant.

### 2. `frontend.Dockerfile` — switch from `FROM scratch` to distroless

Change the final stage to `FROM gcr.io/distroless/static-debian12:nonroot`. The distroless image ships `/tmp` (mode 1777), CA certs, and tzdata out of the box, eliminating an entire class of \"missing rootfs bits\" failures. Attack surface is essentially identical to scratch (no shell, no package manager, ~2 MB). The published image now satisfies `os.TempDir()`'s implicit precondition by default; the Go-side helper above becomes defense-in-depth.

### 3. `test/e2e/frontend/frontend_test.go` — build the real `frontend.Dockerfile`

The previous `buildFrontendImage()` built an inline `FROM alpine:3.18` test-only Dockerfile, which hid the regression because alpine has a working `/tmp`. Replace it with `docker buildx build -f frontend.Dockerfile`. From now on CI exercises the same image release.yml ships, so scratch/distroless-rootfs drift between CI and production becomes impossible.

## Why no test caught this

- Unit tests run on the host, where `/tmp` always exists.
- The e2e test wrapped the binary in alpine (which has `/tmp`) instead of building the real `FROM scratch` Dockerfile, so the production rootfs was never exercised in CI.
- `frontend.Dockerfile` is consumed only by `release.yml`. The first run against a real scratch rootfs was in production, by users.

The new `TestEnsureTempDirCreatesMissingDir` simulates a missing `/tmp` by pointing `TMPDIR` at a non-existent path. Combined with the e2e fidelity fix, both the unit and integration layers now catch this regression class.

## Test plan

- [x] `go test ./pkg/frontend/...` passes (new test included)
- [x] `go vet ./pkg/frontend/... ./test/e2e/frontend/...` clean
- [x] `go build ./...` clean
- [x] Rebuilt `frontend.Dockerfile` (now distroless) and ran a real end-to-end patch against `nginx:1.21` via a Trivy report through `docker buildx` with an `oci-layout` build context. Result: `Installing 60 security updates`, valid `patched.tar` written.
- [x] Independently verified that switching to distroless alone (with the broken `build.go` from `main` un-patched) is also enough to fix the runtime failure — confirms the rootfs is the real precondition.
- [x] Confirmed the original `ghcr.io/project-copacetic/copacetic-frontend:v0.14.0` image fails the same flow with the documented error.

## Bisect

Regression introduced in #1507 (commit `3fa656ac`), \"fix: security hardening from codebase audit.\" The PR description notes \"Fix temp dir creation in frontend/build.go to use os.TempDir() instead of root filesystem,\" which is the right principle but missed the scratch-rootfs precondition.

### DIFF
--- a/frontend.Dockerfile
+++ b/frontend.Dockerfile
@@ -36,9 +36,6 @@ RUN --mount=type=cache,id=gomod,target=/go/pkg/mod \
 # adding those back.
 FROM gcr.io/distroless/static-debian12:nonroot AS frontend
 
-# Copy CA certificates for HTTPS connections
-COPY --from=builder /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
-
 # Copy the frontend binary
 COPY --from=builder /copa-frontend /copa-frontend
 
--- a/frontend.Dockerfile
+++ b/frontend.Dockerfile
@@ -28,7 +28,13 @@ RUN --mount=type=cache,id=gomod,target=/go/pkg/mod \
     CGO_ENABLED=0 GOOS=linux go build -ldflags="-s -w" -o /copa-frontend ./cmd/frontend
 
 # Final image
-FROM scratch AS frontend
+#
+# Distroless static gives us a minimal rootfs (no shell, no package
+# manager) while still providing /tmp, CA certs, /etc/passwd, and
+# tzdata -- which the frontend needs at runtime (e.g. os.MkdirTemp
+# requires /tmp to exist). Do not switch to `FROM scratch` without
+# adding those back.
+FROM gcr.io/distroless/static-debian12:nonroot AS frontend
 
 # Copy CA certificates for HTTPS connections
 COPY --from=builder /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
--- a/pkg/frontend/build.go
+++ b/pkg/frontend/build.go
@@ -24,10 +24,8 @@ const (
 )
 
 // ensureTempDir makes sure the directory returned by os.TempDir() exists
-// before callers invoke os.MkdirTemp("", ...). The published frontend image
-// is built FROM scratch (see frontend.Dockerfile), so /tmp is not present
-// at runtime and MkdirTemp would otherwise fail with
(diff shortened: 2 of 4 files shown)


robert-cronin (approved):
Validated this locally against the actual frontend release path.

Covered:
- `go test ./pkg/frontend/...`
- build of `frontend.Dockerfile`
- rootfs inspection: nonroot image, no shell, `/tmp` present with 1777, CA bundle present from distroless
- real `buildctl --frontend=gateway.v0` single-report patch path
- real `buildctl --frontend=gateway.v0` directory/multiplatform report path
- confirmed published `v0.14.0` frontend reproduces the original `stat /tmp: no such file or directory` failure while this PR image succeeds

I pushed a small cleanup commit to remove stale scratch wording and avoid copying Alpine CA certs over the distroless CA bundle.

PR #1605, opened 2026-05-12
fix: capture original manifest annotations before patching to prevent tag overwrite

jpinz:
This PR resolves an issue for when patching an image using the same tag as the source, the manifest annotations are dropped.

It updates the flow so that the original manifest annotations are saved before patching, then re-applied after patching.


robert-cronin:
Added one commit on top: the lookup timing fix here is correct, but `GetPlatformManifestAnnotations` returned nil for single-platform images, and even when it didn't, the captured annotations only reached the multi-arch manifest assembly, single-platform `--push` never forwarded them to BuildKit's exporter. My commit fixes both, plus factors out the `version` rewrite so single-arch matches multi-arch behavior. E2E verified single + multi-arch.

robert-cronin (approved):

PR #1613, opened 2026-05-19
feat: update github runners

idris-bitvavo:
Update github runners to default


sozercan (changes requested):
why?

sozercan:
Thanks for the pr! We don’t have any plans to move away from current runners at this time.

PR #1614, opened 2026-05-19
fix: additional issues with hitting the registry instead of the local daemon

jpinz:
This pull request enhances how platform and annotation information is discovered for container images by prioritizing local Docker daemon inspection before falling back to remote registry lookups. This makes platform detection and annotation retrieval more robust and enables working with images that exist only locally. The main changes include new utility functions for local image inspection, updates to platform and annotation discovery logic, and improved error handling and logging.


robert-cronin (commented):
The local-daemon fix seems to solve the ACR issue, but there's a bug in the platform discovery code.

`LocalImagePlatforms` and the other helpers don't check `m.Available` when looping through manifests. So when you partially pull a multi-arch image (like the bulk test does), it returns platforms that aren't actually downloaded locally. Then copa tries to patch arm64 without QEMU and fails.

Quick fix: add `if !m.Available { continue }` in the three manifest loops.

The nodejs failure is unrelated, that's a pre-existing Ghost CVE issue.

robert-cronin (approved):

ashnamehrotra (approved):
LGTM

PR #1622, opened 2026-06-03
fix(test/e2e): allowlist known transitive CVEs in nodejs patch verification

robert-cronin:
Unblocks the two CI jobs that are reliably red on every PR against `main`:

**1. `vuln-check` — bump Go to 1.25.11**

`govulncheck` is flagging `GO-2026-5039` (`net/textproto`) and `GO-2026-5037` (`crypto/x509`) in the stdlib, fixed in `1.25.11`. Bumps `go.mod` to `1.25.11` and migrates all workflows from a hardcoded `go-version: "1.25.10"` to `go-version-file: go.mod` so the toolchain version lives in one place (matches the pattern already used by `release.yml`). `frontend.Dockerfile`'s `ARG GO_VERSION` default is bumped too; `release.yml`'s build flow already passes the version through from `go.mod`.

One subtlety in `dependency-review.yml`: `golang/govulncheck-action` defaults `go-version-input: 'stable'` and forwards it to setup-go, which then prefers `go-version` over `go-version-file` when both are set. Cleared the default with `go-version-input: ''` so `go.mod` actually wins.

**2. `Test Node.js patching` — allowlist `CVE-2026-45149`**

`brace-expansion` DoS (CWE-400). Copa's nodejs langmgr upgrades `glob`/`minimatch` to their fixed versions but the resolved transitive `brace-expansion` lands on 5.0.0–5.0.5 instead of the patched 5.0.6. The assertion is doing its job, but blocking every unrelated PR until the langmgr is fixed.

Adds a `knownIntroducedCVEs` allowlist + helper that skips listed CVEs (logged as `KNOWN COPA BUG: ...` so they stay visible) while keeping the canary for any other newly introduced CVE. Does **not** fix the underlying nodejs langmgr gap — follow-up issue to track.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -44,7 +44,7 @@ jobs:
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       - uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.4.0
         with:
-          go-version: "1.25.10"
+          go-version-file: go.mod
           check-latest: true
       - name: Add containerd-snapshotter to docker daemon
         run: |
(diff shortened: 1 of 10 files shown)


omercnet (dismissed):

ashnamehrotra (approved):
LGTM

PR #1634, opened 2026-06-27
fix: detect missing apt-get in chiseled images and fall back to tooling container

AruneshDwivedi:
Some Debian-based images like dotnet/aspnet:8.0-noble-chiseled-extra have a dpkg status file but no apt-get binary. installUpdates was running apt-get update directly on the target image, which fails with "exec: apt-get: executable file not found in $PATH".

Now probeAptGet checks for apt-get availability after detecting DPKGStatusFile. When apt-get is absent, the code sets isDistroless=true and falls back to unpackAndMergeUpdates, which runs apt-get in the tooling container instead of the target image.


AruneshDwivedi:
Thanks for the review. The first point about empty packageInfo is not an issue in this code — probeAptGet sets isDistroless for status-file images but probeDPKGStatus returns nil immediately after, so InstallUpdates never reaches unpackAndMergeUpdates for this status type. The isDistroless flag is only consulted after probeDPKGStatus returns nil. For the second point about the probe command — fixed by adding `|| true` so the shell always exits 0 when apt-get is absent, leaving the marker file absent rather than failing the build step.

robert-cronin:
Thanks for working on this and for helping surface the apt-less Chiseled-image case. It looks like this PR has since been superseded by #1667, which now covers the broader full-status workflow with test coverage on main, so I think we can close this one.

PR #1672, opened 2026-08-16
fix: nil platform panic and OpenVEX schema compliance

### LINKED ISSUE #1645: [BUG] nil platform manifest entry causes a panic in DiscoverPlatformsFromReference
### Version of copa

main built from source

### Expected Behavior

When copa patches a multi platform image whose manifest index contains a child entry that has no platform field, copa should skip that entry and keep discovering the remaining valid platforms.

### Actual Behavior

copa crashes with a nil pointer dereference panic.

The code path is DiscoverPlatformsFromReference in pkg/buildkit/buildkit.go around lines 399 and 400. The branch that is meant to skip entries with a missing or unknown platform still reads m.Platform.OS and m.Platform.Architecture inside the log call. Go evaluates those arguments before the log function runs, and logrus evaluates them regardless of the current log level, so a nil platform is dereferenced and the process panics.

The failing code looks like this

    if m.Platform == nil || m.Platform.OS == "unknown" || m.Platform.Architecture == "unknown" {
        log.Debugf("Skipping manifest with unknown platform: %s/%s", m.Platform.OS, m.Platform.Architecture)
        continue
    }

The panic message is

    runtime error invalid memory address or nil pointer dereference

A manifest index can legitimately contain entries with no platform field, for example attestation or provenance entries, so any such image triggers this.

### Steps To Reproduce

1. Build copa from source on main
2. Prepare a multi platform image whose index has at least one child manifest with no platform field, for example an attestation entry
3. Run copa patch on that image with no report so the platform discovery path runs
4. copa panics instead of skipping the entry and patching the valid platforms

### LINKED ISSUE #1651: [BUG] Generated OpenVEX documents omit required vulnerability.name and fail schema validation
### Version of copa

v0.14.0-40-g00cc7cab

### Expected Behavior

`copa patch --format openvex` should generate a document that conforms to the official OpenVEX JSON Schema.

In particular, every statement's vulnerability object should contain the required `name` property. For example:

```json
{
  "vulnerability": {
    "name": "CVE-2024-28085"
  }
}
```

Copacetic may retain an optional `@id` as appropriate, but it should not use `@id` as a substitute for the required `name`.

This report only concerns the structure and standards conformance of the generated OpenVEX document. It does not assess whether the reported `fixed` statuses are factually correct or whether the patched image remains exploitable.

### Actual Behavior

The generated document contains 285 statements. All 285 vulnerability objects contain only `@id` and omit `name`.

For example:

```json
{
  "vulnerability": {
    "@id": "CVE-2024-28085"
  },
  "products": [
    {
      "@id": "pkg:oci/nginx@sha256:9523316310eb8fbe3441392a6e405630334943f83855b9025f9c4cd66f904aa3"
    }
  ],
  "status": "fixed"
}
```

Validation against the official OpenVEX JSON Schema produces:

```text
Statements: 285
Total schema errors: 285
Missing vulnerability.name errors: 285
Other schema errors: 0

'name' is a required property
Path: /statements/<index>/vulnerability
```

These are 285 occurrences of the same systematic generation issue, not 285 independent root causes.

The OpenVEX specification states that [`name` is the required field in a vulnerability object](https://github.com/openvex/spec/blob/d29fab0c81cfea04159436d54c3edc32947039ac/OPENVEX-SPEC.md#L293-L334), and the [official JSON Schema declares `"required": ["name"]`](https://github.com/openvex/spec/blob/d29fab0c81cfea04159436d54c3edc32947039ac/openvex_json_schema.json#L8-L36).

The current Copacetic implementation initializes only `ID`:

```go
Vulnerability: vex.Vulnerability{ID: u.VulnerabilityID},
```

See [`pkg/vex/openvex.go`](https://github.com/project-copacetic/copacetic/blob/00cc7cabe83945677b9bec24ab71ad02fa5e9a4d/pkg/vex/openvex.go#L130-L137). Copacetic uses `go-vex v0.2.7`, where `Name` is a separate member with `json:"name,omitempty"`. Because Copacetic does not set it, serialization omits it.

The current [unit-test fixture](https://github.com/project-copacetic/copacetic/blob/00cc7cabe83945677b9bec24ab71ad02fa5e9a4d/pkg/vex/openvex_test.go#L43-L95) and [documentation example](https://github.com/project-copacetic/copacetic/blob/00cc7cabe83945677b9bec24ab71ad02fa5e9a4d/website/docs/output.md#L35-L47) also contain ID-only vulnerability objects. The [end-to-end test](https://github.com/project-copacetic/copacetic/blob/00cc7cabe83945677b9bec24ab71ad02fa5e9a4d/test/e2e/generate/generate_test.go#L321-L337) checks basic output fields but does not validate the document against the official schema.

I searched the public Copacetic issues and pull requests for `vulnerability.name`, the exact schema error, and invalid OpenVEX output, but did not find an existing report that directly addresses this omission. Issues [#667](https://github.com/project-copacetic/copacetic/issues/667) and [#1536](https://github.com/project-copacetic/copacetic/issues/1536) concern different VEX identifier and BOM-correlation problems.

## Additional non-blocking observations

The following are interoperability recommendations, not schema errors and not requirements for resolving the bug above:

- The generated document uses `"@context": "https://openvex.dev/ns"`. The [OpenVEX specification](https://github.com/openvex/spec/blob/d29fab0c81cfea04159436d54c3edc32947039ac/OPENVEX-SPEC.md#L150-L164) describes a versioned form such as `https://openvex.dev/ns/v0.2.0`. The current unversioned value still satisfies the JSON Schema's URI constraint.
- The generated document uses `"author": "Project Copacetic"`. This identifies an organization and satisfies the required author field. The specification says that the author should ideally be represented by a machine-readable identifier such as an IRI or email address, but the schema does not require this. Copacetic also allows users to override the value with `COPA_VEX_AUTHOR`.

### Attachments

[copa_openvex.json](https://github.com/user-attachments/files/30173693/copa_openvex.json) — generated OpenVEX output
[trivy_for_copa.json](https://github.com/user-attachments/files/30173704/trivy_for_copa.json) — Trivy input report

### Steps To Reproduce

1. Prepare the local container image `nginx:1.21.6` for `linux/arm64` (Debian 11.3).
2. Generate the attached `trivy_for_copa.json` using Trivy `0.69.3`; the vulnerability DB was updated at `2026-07-20T01:09:07Z`.
3. Run `copa patch --image nginx:1.21.6 --report <report-directory>/trivy_for_copa.json --tag 1.21.6-copa-patched --loader docker --timeout 30m --format openvex --output <output-directory>/copa_openvex.json --progress plain`, replacing the placeholder directories with the desired local paths.

AruneshDwivedi:
DiscoverPlatformsFromReference panicked when m.Platform was nil because the debug log call read m.Platform.OS and m.Platform.Architecture unconditionally inside the guard. The guard short-circuits on m.Platform==nil but Go still evaluates both sides of the || before the log call.

Also populate Vulnerability.Name in generated OpenVEX statements so documents pass schema validation. The go-vex library requires name as a mandatory field.

Fixes #1645
Fixes #1651


omercnet (dismissed):

pkg/buildkit/buildkit.go
```diff
@@ -433,6 +433,34 @@ func DiscoverPlatformsFromReference(manifestRef string) ([]types.PatchPlatform,
 			return nil, fmt.Errorf("error getting manifest: %w", err)
 		}
 
+		for i := range manifest.Manifests {
```
robert-cronin: The branch already contains `platformsFromIndexManifest` from `main`, including nil-platform handling and its regression test. This duplicated loop returns before the helper call, leaving that call unreachable (`go vet ./pkg/buildkit` fails) and disconnecting the regression test from the production path. Please remove this loop and return `platformsFromIndexManifest(manifest)` directly.

pkg/vex/openvex.go
```diff
@@ -131,7 +131,7 @@ func (o *OpenVex) createVEXDocument(
 		product := imageProduct
 		product.Subcomponents = []vex.Subcomponent{subComponent}
 		doc.Statements = append(doc.Statements, vex.Statement{
-			Vulnerability: vex.Vulnerability{ID: u.VulnerabilityID},
+			Vulnerability: vex.Vulnerability{Name: vex.VulnerabilityID(u.VulnerabilityID), ID: u.VulnerabilityID},
```
robert-cronin, line 134: Could you also update `website/docs/output.md`? Its OpenVEX example still shows the old ID-only vulnerability object, so the documented output remains schema-invalid after this user-visible format change.

AruneshDwivedi (commented):
.

robert-cronin (approved):

ashnamehrotra (approved):

PR #1682, opened 2026-08-24
fix(langmgr): append +incompatible suffix for pre-module major>=2 Go deps

omercnet:
## Problem

`go get` fails outright when patching a pre-modules-era Go dependency released at major version 2 or higher whose module path carries no `/vN` suffix. The canonical case is `github.com/docker/docker@v28.x`: the module path is not modules-aware, so the Go toolchain only accepts the version with the `+incompatible` build tag. Scanner reports routinely report the bare version (`v28.0.0`), and the patch run dies on the `go get` step.

## Root cause

`pkg/langmgr/golang.go` builds the `go get` spec string verbatim from the report's fixed version in two places, `updateGoModule` and `upgradePackagesWithTooling`. Both only normalize the `v` prefix, so a bare `v28.0.0` is passed through unchanged.

## Fix

New helper `appendIncompatibleIfNeeded(modulePath, version)` applied at both spec construction sites, right after the `v`-prefix normalization. It appends `+incompatible` only when all of the following hold:

- the version is valid semver (non-semver input is returned untouched)
- it is not already suffixed
- the major is not v0 or v1
- `module.SplitPathVersion` reports no path major suffix, meaning the dependency is not modules-aware

Everything else is returned unchanged, so `/v2`-style modules, v0/v1 deps, and pseudo-versions below v2 keep their current behavior.

## Tests

- `TestAppendIncompatibleIfNeeded`: table-driven coverage for `github.com/docker/docker@v28.0.0`, `/v2` path suffix, v0 and v1 majors, already-suffixed input, pseudo-versions at v0 and at v2 with and without a path suffix, and invalid/empty version input.
- `TestIncompatibleVersionsPassValidation`: confirms `isValidGoVersion`, `validateGoVersion`, and `cleanGoVersion` all accept a `+incompatible` version (semver treats it as build metadata, and `+` is not in the shell-unsafe character set).

`go build ./...`, `go test ./pkg/langmgr`, and `gofumpt` all pass; `make lint` reports only pre-existing findings unrelated to this change.

### DIFF
--- a/pkg/langmgr/golang.go
+++ b/pkg/langmgr/golang.go
@@ -13,6 +13,7 @@ import (
 	"github.com/project-copacetic/copacetic/pkg/types/unversioned"
 	"github.com/project-copacetic/copacetic/pkg/utils"
 	log "github.com/sirupsen/logrus"
+	"golang.org/x/mod/module"
 	"golang.org/x/mod/semver"
 )
 
@@ -174,6 +175,30 @@ func cleanGoVersion(version string) string {
 	}
 
 	return ""
+}
+
+// appendIncompatibleIfNeeded adds the "+incompatible" build tag that `go get`
+// requires for pre-modules dependencies released at major version 2 or higher
+// whose module path carries no /vN suffix (github.com/docker/docker@v28.0.0,
+// for example). Scanner reports commonly omit the tag and `go get` rejects the
+// bare version outright.
+func appendIncompatibleIfNeeded(modulePath, version string) string {
+	if !semver.IsValid(version) || strings.HasSuffix(version, "+incompatible") {
+		return version
+	}
+
+	switch semver.Major(version) {
+	case "v0", "v1":
+		return version
+	}
+
+	// A path major suffix (/v2, /v3, ...) means the dependency is modules-aware,
+	// so the version is used as-is.
+	if _, pathMajor, ok := module.SplitPathVersion(modulePath); !ok || pathMajor != "" {
+		return version
+	}
+
+	return version + "+incompatible"
 }
 
 // buildGoUpdateCmd assembles the shell command run inside the build container
@@ -942,6 +967,7 @@ func (gm *golangManager) updateGoModule(
 			if !strings.HasPrefix(version, "v") {
 				version = "v" + version
 			}
+			version = appendIncompatibleIfNeeded(u.Name, version)
 			spec := fmt.Sprintf("%s@%s", u.Name, version)
 			getCommands = append(getCommands, fmt.Sprintf("go get %s", spec))
 		} else {
@@ -1069,6 +1095,7 @@ func (gm *golangManager) upgradePackagesWithTooling(
 				if !strings.HasPrefix(version, "v") {
 					version = "v" + version
 				}
+				version = appendIncompatibleIfNeeded(u.Name, version)
 				spec := fmt.Sprintf("%s@%s", u.Name, version)
 				getCommands = append(getCommands, fmt.Sprintf("go get %s", spec))
 			}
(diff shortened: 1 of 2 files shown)


pkg/langmgr/golang.go
```diff
@@ -910,6 +935,7 @@ func (gm *golangManager) updateGoModule(
 			if !strings.HasPrefix(version, "v") {
 				version = "v" + version
 			}
+			version = appendIncompatibleIfNeeded(u.Name, version)
```
robert-cronin, line 985: Could we normalize the binary-rebuild update map too, since it still writes bare v2+ requirements?

pkg/langmgr/golang.go
```diff
@@ -153,6 +154,30 @@ func cleanGoVersion(version string) string {
...
+		return version
+	}
+
+	// A path major suffix (/v2, /v3, ...) means the dependency is modules-aware,
+	// so the version is used as-is.
+	if _, pathMajor, ok := module.SplitPathVersion(modulePath); !ok || pathMajor != "" {
+		return version
+	}
+
+	return version + "+incompatible"
```
robert-cronin, line 203: Could this handle an existing build suffix before adding `+incompatible`, so it never produces two `+` components?

pkg/langmgr/golang.go
omercnet: Done: the binary-rebuild path now builds its update map via `buildBinaryUpdateMap`, which applies the `v` prefix and `appendIncompatibleIfNeeded` before the versions reach the go.mod requirement edits, so it no longer writes bare v2+ requirements. `TestBuildBinaryUpdateMap` covers that construction, including a bare `28.0.0` input and a `/v2` module path that must stay untagged.

pkg/langmgr/golang.go
omercnet: Good catch: the guard is now `strings.Contains(version, "+")` instead of a literal `+incompatible` suffix check, so any existing build metadata short-circuits and we can never emit two `+` components. Added a regression case asserting `v2.0.0+build1` is returned unchanged.

omercnet:
Review feedback addressed in cb82a16 (rebased on current main):

- The binary-rebuild path now builds its update map through `buildBinaryUpdateMap`, which applies the `v` prefix and `appendIncompatibleIfNeeded` before the versions become go.mod requirements, so it no longer records bare pre-modules v2+ requirements.
- `appendIncompatibleIfNeeded` now short-circuits on any existing build metadata (`strings.Contains(version, "+")`) rather than only a literal `+incompatible` suffix, so `v2.0.0+build1` can never grow a second `+` component.
- Regression coverage: `TestBuildBinaryUpdateMap` and a new `v2.0.0+build1` case in `TestAppendIncompatibleIfNeeded`.

CI: 74 of 76 checks pass. The two failures, `lint` and `Test Bulk Patching`, are inherited from main and unrelated to this diff, which only touches `pkg/langmgr`. Both come from `integration/bulk/singlearch_patch_test.go` failing typecheck after the testcontainers 0.44.0 bump (`docker/docker` vs `moby/moby` container types); the same two checks fail on main's head. #1685 fixes that, and this PR should go green once it lands.

robert-cronin (dismissed):
Both findings are resolved; could you rebase onto current main so the #1683 tests and this PR's tests are both retained and CI reruns?

omercnet:
Rebased onto current main (`2a7738b`, which includes #1683) and force-pushed as `7b5b429`. Note this dismissed the approval, so a re-approve is needed once CI settles.

The conflict was in `pkg/langmgr/golang_test.go`, where both branches appended test functions at the same spot. Both sets are retained in full: `TestFilterGoDowngrades` and `TestGolangManagerInstallUpdatesSkipsNonNewerVersion` from #1683, plus `TestAppendIncompatibleIfNeeded`, `TestIncompatibleVersionsPassValidation` and `TestBuildBinaryUpdateMap` from this PR. All five pass, and the diff against main is still confined to `pkg/langmgr` with no deletions from #1683.

robert-cronin (approved):

PR #1683, opened 2026-08-24
fix(langmgr): skip Go updates where fixed version is not newer than installed

omercnet:
## Problem

The Go language manager can downgrade a dependency. When a scanner report carries a stale advisory, or the image has already been patched to a newer version, `copa` runs `go get pkg@version` with the reported fixed version even though the installed version is already newer or equal.

## Root cause

`GetUniqueLatestUpdates` (`pkg/langmgr/langmgr.go`) deduplicates update requests and picks the highest fixed version per module. It does not compare that version with `InstalledVersion`. `golangManager.InstallUpdates` (`pkg/langmgr/golang.go`) therefore applies the reported `FixedVersion` without checking whether it is newer.

## Fix

Add `filterGoDowngrades` after update deduplication and validation. It skips an update with a warning when both versions are present and valid Go semantic versions, and the installed version is greater than or equal to the fixed version.

Missing or unparsable versions retain the filter's previous pass-through behavior. If every module update is filtered and no standard-library update remains, `InstallUpdates` returns the current state without constructing an upgrade.

A skipped downgrade leaves the module untouched in the image, so `filterGoDowngrades` returns the skipped module names and `InstallUpdates` adds them to `errPkgsReported`. That keeps them out of the validated updates built in `pkg/patch`, so the VEX document does not claim the finding was remediated when nothing changed. Consistent with existing `errPkgsReported` entries, this reports the package as unpatched and is not fatal on its own.

## Test coverage

- `TestFilterGoDowngrades` covers older, equal, newer, unprefixed, missing, unparsable, and mixed version inputs, and asserts that every skipped module is returned as skipped and absent from the updates that get applied.
- `TestGolangManagerInstallUpdatesSkipsNonNewerVersion` covers the manager-level early return, verifying the current state is preserved and the skipped module is reported as unpatched.

## Note on the second commit

`fix(test): align bulk integration test with moby container types` is unrelated to the Go downgrade guard. It repairs a build break already present on `main`: after the testcontainers-go bump, `integration/bulk/singlearch_patch_test.go` no longer compiles against `docker/docker` container types, which fails both `lint` (typecheck) and `Test Bulk Patching`. It is a separate commit so it can be dropped or cherry-picked independently.

## Verification

- `go test ./pkg/langmgr ./pkg/patch -count=1`
- `go build ./...`
- `go vet ./integration/...`
- `gofumpt -l pkg/langmgr integration/bulk` (clean)

### DIFF
--- a/pkg/langmgr/golang.go
+++ b/pkg/langmgr/golang.go
@@ -128,6 +128,24 @@ func isLessThanGoVersion(v1, v2 string) bool {
 	return semver.Compare(v1, v2) < 0
 }
 
+// filterGoDowngrades drops updates whose reported fixed version is not newer than
+// the installed version. A stale advisory would otherwise downgrade a module that
+// has already been patched. Updates with missing or unparsable versions are kept
+// so that existing behavior is unchanged.
+func filterGoDowngrades(updates unversioned.LangUpdatePackages) unversioned.LangUpdatePackages {
+	filtered := make(unversioned.LangUpdatePackages, 0, len(updates))
+	for _, u := range updates {
+		if u.InstalledVersion != "" && u.FixedVersion != "" &&
+			isValidGoVersion(u.InstalledVersion) && isValidGoVersion(u.FixedVersion) &&
+			!isLessThanGoVersion(u.InstalledVersion, u.FixedVersion) {
+			log.Warnf("Skipping Go module %s: installed version %s is not older than fixed version %s", u.Name, u.InstalledVersion, u.FixedVersion)
+			continue
+		}
+		filtered = append(filtered, u)
+	}
+	return filtered
+}
+
 // cleanGoVersion extracts the first valid version from a comma-separated list.
 // This handles cases where Trivy returns multiple versions.
 func cleanGoVersion(version string) string {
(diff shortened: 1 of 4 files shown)


pkg/langmgr/golang.go
```diff
@@ -284,6 +302,12 @@ func (gm *golangManager) InstallUpdates(
 		}
 	}
 
+	updatesToAttempt = filterGoDowngrades(updatesToAttempt)
+	if len(updatesToAttempt) == 0 && !hasStdlib {
+		log.Warn("No Go update packages remain to apply after filtering out non-newer fixed versions.")
+		return currentState, errPkgsReported, nil
```
robert-cronin, line 316: Please keep skipped downgrades out of validated updates; this leaves the image unchanged but VEX later marks the finding fixed.

omercnet: Good catch. filterGoDowngrades now returns the skipped module names and InstallUpdates appends them to errPkgsReported, so a skipped downgrade is reported as unpatched and gets excluded from the validated updates that feed the VEX document. Tests cover both halves: the skipped package shows up in the reported packages and is absent from the updates that get applied.

robert-cronin (approved):

PR #1684, opened 2026-08-24
fix(langmgr): verify go.mod reflects applied Go module updates

omercnet:
## Problem

Both Go module update paths in `pkg/langmgr/golang.go` (`updateGoModule` for the in-image toolchain and `upgradePackagesWithTooling` for the tooling container) run `go get <mod>@<ver> && go mod tidy -e` and treated a zero exit code as proof the bump landed. There was no post-update check on the resulting `go.mod` anywhere in the Go path.

## Root cause

`go mod tidy` removes requirements for modules that are not imported by reachable code, and `-e` makes it continue past load errors while doing so. A `go get` bump could therefore be reverted by the tidy step in the same shell command while copa still reported the package as patched.

## Fix

- Add `verifyGoModUpdates`, which parses post-update `go.mod` metadata and uses full-module `go.sum` entries for pre-Go 1.17 indirect dependencies, failing when a requested module is absent, unprovable, or below its fixed version.
- Resolve Go replacement semantics conservatively:
  - an applicable versioned replacement overrides the selected requirement for the old module identity;
  - version-qualified replacements apply only when their old version matches the selected requirement;
  - a replacement to a different module path, or to local unversioned contents, fails verification because module metadata cannot prove which source version it contains;
  - `go.work` replacements are resolved first and override the member module's own `replace` directives, as the go command does.
- Extract and verify the updated `go.mod` after both the in-image and tooling-container update steps.
- Resolve `go.work` members and verify each member module because the workspace root has no authoritative `go.mod`.
- Scope verification per module: each module is only accountable for the updates it already referenced, and a package counts as proven only once every accountable module has been checked. Proving a module in one workspace member no longer excuses another member that still requires the vulnerable version.
- Report every unproven package on failure. Module accountability is collected before any module is touched, so a failure partway through the tooling-container loop still reports the packages carried by the modules queued behind it instead of letting `--ignore-errors` claim them as remediated.
- Limit the `go.sum` fallback to modules whose `go` directive predates module graph pruning (absent or below 1.17). From Go 1.17 on, `go.mod` lists every dependency of the build, so a `go.sum` entry it never mentions is a stale checksum and would otherwise demand a bump from a module that does not use the package.

Existing `--ignore-errors` handling is unchanged. The caller loop in `upgradePackages` still decides whether a per-module verification error is fatal.

## Tests

`TestVerifyGoModUpdates` covers target, missing, below-target, above-target, no-`v` prefix, `+incompatible`, malformed, multi-module, pre-Go 1.17 `go.sum`, unprovable `go.mod` replacement, and `go.work` replacement precedence cases. `TestFilterUpdatesInGoMod` covers the pruned-module-graph gate on `go.sum`, and `TestParseGoWorkspace` covers member resolution plus workspace replacement extraction. Behavior-level regressions run `updateGoModule` and `upgradePackagesWithTooling` against a fake BuildKit gateway:

- `TestUpdateGoModuleWorkspaceProofIsPerMember` — a module proven in one member is still reported when another member owes proof.
- `TestUpdateGoModuleWorkspaceReplacementIsUnprovable` — a `go.work` local replacement fails verification.
- `TestUpgradePackagesWithToolingReportsUnprocessedModules` — a failure reports packages of modules never processed.
- `TestUpdateGoModuleWorkspaceFailureReportsAllUnverified` — a read failure reports every update lacking proof.

Existing coverage for `+incompatible` normalization from #1682 and downgrade/VEX behavior from #1683 is preserved.

Verified locally with:

- `make format`
- `go build ./...`
- `go vet ./pkg/langmgr`
- `go test ./pkg/langmgr ./pkg/patch`

### DIFF
--- a/pkg/langmgr/golang.go
+++ b/pkg/langmgr/golang.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"path"
 	"strings"
 
 	"github.com/moby/buildkit/client/llb"
@@ -218,8 +219,9 @@ func buildGoUpdateCmd(modPath, allGetCmd string) string {
 // the go.mod produced by the update step. `go mod tidy -e` removes requirements
 // for modules that are no longer imported by reachable code, so a successful
 // `go get` can be silently reverted and the package still reported as patched.
-// Applicable versioned replace directives override the old module's require
-// version and also count as an effective instance of their replacement target.
+// Only a replace directive whose left-hand side is the module under test can
+// change that module's effective version: versions of other module paths are
+// not comparable to the requested one.
 func verifyGoModUpdates(goModContent []byte, updates unversioned.LangUpdatePackages) error {
 	hasTargets := false
 	for _, u := range updates {
@@ -249,50 +251,109 @@ func verifyGoModUpdates(goModContent []byte, updates unversioned.LangUpdatePacka
 			continue
 		}
 
-		found := false
-		if requiredVersion, ok := required[u.Name]; ok {
-			found = true
-			landedVersion := requiredVersion
-			for _, rep := range f.Replace {
-				if rep == nil || rep.Old.Path != u.Name {
-					continue
-				}
-				if rep.Old.Version != "" && rep.Old.Version != requiredVersion {
-					continue
-				}
-				if rep.New.Version == "" {
-					return fmt.Errorf("module %s has local replacement %s; cannot verify requested version %s", u.Name, rep.New.Path, u.FixedVersion)
-				}
-				landedVersion = rep.New.Version
-				break
-			}
-			if isLessThanGoVersion(landedVersion, u.FixedVersion) {
-				return fmt.Errorf("module %s is at %s in go.mod after update, expected %s or newer", u.Name, landedVersion, u.FixedVersion)
-			}
-		}
-
-		// A module can also enter the build list as the target of a replacement
-		// for another required module.
-		for _, rep := range f.Replace {
-			if rep == nil || rep.Old.Path == u.Name || rep.New.Path != u.Name || rep.New.Version == "" {
-				continue
-			}
-			oldVersion, ok := required[rep.Old.Path]
-			if !ok || (rep.Old.Version != "" && rep.Old.Version != oldVersion) {
-				continue
-			}
-			found = true
-			if isLessThanGoVersion(rep.New.Version, u.FixedVersion) {
-				return fmt.Errorf("module %s is at %s in go.mod after update, expected %s or newer", u.Name, rep.New.Version, u.FixedVersion)
-			}
-		}
-
-		if !found {
+		requiredVersion, ok := required[u.Name]
+		if !ok {
 			return fmt.Errorf("module %s not found in go.mod after update; the requested version %s was not applied or was removed by 'go mod tidy'", u.Name, u.FixedVersion)
+		}
+
+		landedVersion := requiredVersion
+		if replacement := effectiveReplacementVersion(f.Replace, u.Name, requiredVersion); replacement != "" {
+			landedVersion = replacement
+		}
+		if isLessThanGoVersion(landedVersion, u.FixedVersion) {
+			return fmt.Errorf("module %s is at %s in go.mod after update, expected %s or newer", u.Name, landedVersion, u.FixedVersion)
 		}
 	}
 
 	return nil
+}
+
+// effectiveReplacementVersion returns the version that replace directives give
+// modPath at selectedVersion, or "" when no directive names a version for it.
+// Mirroring go.mod resolution, a directive pinned to the selected version wins
+// over a path-wide directive, and a filesystem replacement (which carries no
+// version) never overrides the require entry.
+func effectiveReplacementVersion(replacements []*modfile.Replace, modPath, selectedVersion string) string {
+	var pinned, pathWide string
+	for _, rep := range replacements {
+		if rep == nil || rep.Old.Path != modPath || rep.New.Version == "" {
+			continue
+		}
+		switch {
+		case rep.Old.Version == "":
+			if pathWide == "" {
+				pathWide = rep.New.Version
+			}
+		case rep.Old.Version == selectedVersion:
+			if pinned == "" {
+				pinned = rep.New.Version
+			}
+		}
+	}
+	if pinned != "" {
+		return pinned
+	}
+	return pathWide
+}
+
+// filterUpdatesInGoMod narrows updates to the modules the given go.mod already
+// referenced. The update step runs the same `go get` list against every module
+// in the image, so verifying all of them against one go.mod would fail modules
+// that never depended on the vulnerable package. Callers pass the module's
+// pre-update go.mod so that requirements dropped by the update are still
+// verified rather than silently excused.
+func filterUpdatesInGoMod(goModContent []byte, updates unversioned.LangUpdatePackages) (unversioned.LangUpdatePackages, error) {
+	f, err := modfile.Parse("go.mod", goModContent, nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse go.mod: %w", err)
+	}
+
+	present := make(map[string]struct{}, len(f.Require)+len(f.Replace))
+	for _, r := range f.Require {
+		if r != nil && r.Mod.Path != "" {
+			present[r.Mod.Path] = struct{}{}
+		}
+	}
+	for _, rep := range f.Replace {
+		if rep != nil && rep.Old.Path != "" {
+			present[rep.Old.Path] = struct{}{}
+		}
+	}
+
+	var applicable unversioned.LangUpdatePackages
+	for _, u := range updates {
+		if _, ok := present[u.Name]; ok {
+			applicable = append(applicable, u)
+		}
+	}
+	return applicable, nil
+}
+
+// goWorkspaceMemberModPaths resolves the go.mod path of every module listed by a
+// go.work file, relative to the workspace root. A workspace root has no
+// authoritative go.mod of its own: requirements land in the member modules, so
+// those are the files to verify.
+func goWorkspaceMemberModPaths(goWorkContent []byte, workRoot string) ([]string, error) {
+	f, err := modfile.ParseWork("go.work", goWorkContent, nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse go.work: %w", err)
+	}
+
+	var modPaths []string
+	for _, use := range f.Use {
+		if use == nil || use.Path == "" {
+			continue
+		}
+		dir := use.Path
+		if !path.IsAbs(dir) {
+			dir = path.Join(workRoot, dir)
+		}
+		modPaths = append(modPaths, path.Join(path.Clean(dir), "go.mod"))
+	}
+	if len(modPaths) == 0 {
+		return nil, fmt.Errorf("go.work at %s lists no module directories; cannot verify Go module updates", workRoot)
+	}
+	return modPaths, nil
 }
 
 // filterGoPackages filters for Go module and binary packages.
@@ -645,13 +706,14 @@ func (gm *golangManager) upgradePackages(
 	// If workspace exists, update from workspace root
 	if workspacePath != "" {
 		log.Debugf("Updating Go workspace at %s", workspacePath)
-		state, err = gm.updateGoModule(ctx, &state, workspacePath, updates, true, ignoreErrors)
+		var modFailedPkgs []string
+		state, modFailedPkgs, err = gm.updateGoModule(ctx, &state, workspacePath, updates, true, ignoreErrors)
 		if err != nil {
 			log.Errorf("Failed to update workspace: %v", err)
+			// Report the affected packages either way: with errors ignored they
+			// must still surface as unpatched instead of being claimed as fixed.
+			failedPackages = append(failedPackages, modFailedPkgs...)
 			if !ignoreErrors {
-				for _, u := range updates {
-					failedPackages = append(failedPackages, u.Name)
-				}
 				return currentState, failedPackages, err
 			}
 		}
@@ -659,13 +721,11 @@ func (gm *golangManager) upgradePackages(
 		// Update each module independently
 		for _, modPath := range goModPaths {
 			log.Debugf("Updating Go module at %s", modPath)
-			newState, modErr := gm.updateGoModule(ctx, &state, modPath, updates, false, ignoreErrors)
+			newState, modFailedPkgs, modErr := gm.updateGoModule(ctx, &state, modPath, updates, false, ignoreErrors)
 			if modErr != nil {
 				log.Errorf("Failed to update module at %s: %v", modPath, modErr)
+				failedPackages = append(failedPackages, modFailedPkgs...)
 				if !ignoreErrors {
-					for _, u := range updates {
-						failedPackages = append(failedPackages, u.Name)
-					}
 					return currentState, failedPackages, modErr
 				}
 				continue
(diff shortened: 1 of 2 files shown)


pkg/langmgr/golang.go
```diff
@@ -163,6 +164,87 @@ func buildGoUpdateCmd(modPath, allGetCmd string) string {
...
+				if rep == nil || rep.Old.Path != u.Name {
+					continue
+				}
+				if rep.Old.Version != "" && rep.Old.Version != requiredVersion {
+					continue
+				}
+				if rep.New.Version == "" {
+					return fmt.Errorf("module %s has local replacement %s; cannot verify requested version %s", u.Name, rep.New.Path, u.FixedVersion)
+				}
+				landedVersion = rep.New.Version
```
robert-cronin: Versions from different module paths aren't comparable, so this can accept an unrelated replacement as patched.

pkg/langmgr/golang.go
```diff
@@ -163,6 +164,87 @@ func buildGoUpdateCmd(modPath, allGetCmd string) string {
...
+				break
```
robert-cronin: Could this prefer an exact-version replacement over the wildcard, matching Go's replacement precedence?

pkg/langmgr/golang.go
```diff
@@ -939,6 +1021,19 @@ func (gm *golangManager) updateGoModule(
 		llb.WithProxy(utils.GetProxy()),
 	).Root()
 
+	// Confirm the bumps survived 'go mod tidy -e' instead of trusting the exit
+	// code. Skipped for workspaces: the go.work root has no authoritative
+	// go.mod, since requirements land in the member modules.
+	if !isWorkspace {
```
robert-cronin: Could we verify each `go.work` member instead of skipping workspace verification?

pkg/langmgr/golang.go
```diff
@@ -939,6 +1021,19 @@ func (gm *golangManager) updateGoModule(
...
+		goModBytes, err := buildkit.ExtractFileFromState(ctx, gm.config.Client, &state, modPath+"/go.mod")
+		if err != nil {
+			return state, fmt.Errorf("failed to read %s/go.mod after update: %w", modPath, err)
+		}
+		if err := verifyGoModUpdates(goModBytes, updates); err != nil {
```
robert-cronin: Could we verify only updates applicable to this module path before failing on a tidy removal?

pkg/langmgr/golang.go
```diff
@@ -1058,6 +1153,15 @@ func (gm *golangManager) upgradePackagesWithTooling(
 			llb.WithProxy(utils.GetProxy()),
 		).Root()
 
+		// Confirm the bumps survived 'go mod tidy -e' instead of trusting the exit code.
+		goModBytes, extractErr := buildkit.ExtractFileFromState(ctx, gm.config.Client, &toolingState, "/workspace/go.mod")
+		if extractErr != nil {
+			return currentState, nil, fmt.Errorf("failed to read go.mod after update at %s: %w", modPath, extractErr)
+		}
+		if verifyErr := verifyGoModUpdates(goModBytes, updates); verifyErr != nil {
+			return currentState, nil, fmt.Errorf("go module update verification failed at %s: %w", modPath, verifyErr)
```
robert-cronin: Please return these updates as failed so `--ignore-errors` cannot include them in the validated/VEX set.

pkg/langmgr/golang.go
omercnet: Fixed. A replace directive now only changes the effective version of the module named on its left-hand side, so an unrelated module's replacement can no longer stand in for the target; the reverse lookup keyed on `rep.New.Path` is gone. `TestVerifyGoModUpdates` covers the unrelated-replacement case explicitly.

pkg/langmgr/golang.go
omercnet: Added, in `effectiveReplacementVersion`: a directive pinned to the selected version wins over a path-wide one, and a filesystem replacement (which carries no version) never overrides the require entry it falls back to. Both orderings have table cases.

pkg/langmgr/golang.go
omercnet: Workspaces are verified now instead of skipped. `go.work` is parsed with `modfile.ParseWork`, each `use` entry resolves to that member's `go.mod`, and every member is verified against only the updates it referenced. `goWorkspaceMemberModPaths` has unit coverage for relative, block, absolute, and empty `use` forms.

pkg/langmgr/golang.go
omercnet: Each module's `go.mod` is now captured before the update step, and `filterUpdatesInGoMod` restricts verification to updates whose module path was already in that require/replace set. Applied to both the in-image path and this tooling container path.

pkg/langmgr/golang.go
omercnet: Verification failures now return the affected package names, and `upgradePackages` records them whether or not errors are ignored, so `--ignore-errors` reports them unpatched and `getValidatedUpdates` keeps them out of the validated/VEX set. `TestUpdateGoModuleReportsFailedPackages` locks that contract in.

omercnet:
Review feedback addressed in 284ee7f: replacement resolution scoped to the module under verification with go.mod precedence, workspace members verified via `go.work` `use` entries, verification scoped to each module's pre-update require/replace set, and verification failures now returned as failed packages so `--ignore-errors` cannot feed them into the validated/VEX set.

Heads up on CI: `lint` and `Test Bulk Patching` fail here for a reason unrelated to this branch. testcontainers-go 0.44.0 (#1568) broke `integration/bulk/singlearch_patch_test.go` compilation on main, so both checks are red on main itself and on every open PR. Fix is up in #1685; this branch goes green once that lands and this one is rebased.

pkg/langmgr/golang.go
```diff
@@ -163,6 +165,147 @@ func buildGoUpdateCmd(modPath, allGetCmd string) string {
...
+			continue
+		}
+		switch {
+		case rep.Old.Version == "":
+			if pathWide == "" {
+				pathWide = rep.New.Version
+			}
+		case rep.Old.Version == selectedVersion:
+			if pinned == "" {
+				pinned = rep.New.Version
```
robert-cronin: Versions from different module paths still aren't comparable here, so this can accept an unrelated fork as patched.

pkg/langmgr/golang.go
```diff
@@ -163,6 +165,147 @@ func buildGoUpdateCmd(modPath, allGetCmd string) string {
...
+
+// effectiveReplacementVersion returns the version that replace directives give
+// modPath at selectedVersion, or "" when no directive names a version for it.
+// Mirroring go.mod resolution, a directive pinned to the selected version wins
+// over a path-wide directive, and a filesystem replacement (which carries no
+// version) never overrides the require entry.
+func effectiveReplacementVersion(replacements []*modfile.Replace, modPath, selectedVersion string) string {
+	var pinned, pathWide string
+	for _, rep := range replacements {
+		if rep == nil || rep.Old.Path != modPath || rep.New.Version == "" {
```
robert-cronin: A local replacement's contents are unversioned, so the require version cannot prove that code is patched.

pkg/langmgr/golang.go
```diff
@@ -163,6 +165,147 @@ func buildGoUpdateCmd(modPath, allGetCmd string) string {
...
+	}
+	for _, rep := range f.Replace {
+		if rep != nil && rep.Old.Path != "" {
+			present[rep.Old.Path] = struct{}{}
+		}
+	}
+
+	var applicable unversioned.LangUpdatePackages
+	for _, u := range updates {
+		if _, ok := present[u.Name]; ok {
```
robert-cronin, line 411: Could this include pre-1.17 go.sum-only dependencies that Trivy reports for the module?

pkg/langmgr/golang.go
```diff
@@ -939,6 +1139,18 @@ func (gm *golangManager) updateGoModule(
 		llb.WithProxy(utils.GetProxy()),
 	).Root()
 
+	// Confirm the bumps survived 'go mod tidy -e' instead of trusting the exit
+	// code.
+	for _, target := range verifyTargets {
+		goModBytes, extractErr := buildkit.ExtractFileFromState(ctx, gm.config.Client, &state, target.goModPath)
+		if extractErr != nil {
+			return state, getPackageNames(target.updates), fmt.Errorf("failed to read %s after update: %w", target.goModPath, extractErr)
```
robert-cronin: Could a workspace solve failure report every still-unverified update instead of only this target's packages?

pkg/langmgr/golang.go
```diff
@@ -163,6 +165,147 @@ func buildGoUpdateCmd(modPath, allGetCmd string) string {
...
+// Mirroring go.mod resolution, a directive pinned to the selected version wins
+// over a path-wide directive, and a filesystem replacement (which carries no
+// version) never overrides the require entry.
+func effectiveReplacementVersion(replacements []*modfile.Replace, modPath, selectedVersion string) string {
+	var pinned, pathWide string
+	for _, rep := range replacements {
+		if rep == nil || rep.Old.Path != modPath || rep.New.Version == "" {
+			continue
+		}
+		switch {
```
robert-cronin: Could this use a tagged switch so the changed package passes staticcheck QF1002?

pkg/langmgr/golang.go
omercnet: Right, comparing across module paths was still wrong. A replacement whose right-hand side names a different module path is now reported as unverifiable rather than compared: `verifyGoModUpdates` fails with "a different module path cannot prove that requested version was applied". Covered by the `cross-module replacement cannot prove requested version` case.

pkg/langmgr/golang.go
omercnet: Agreed, the require version proves nothing about local contents. The fallback is gone: a filesystem replacement governing the module now fails verification as unversioned, so the package is reported unpatched instead of assumed fixed. Covered by the `filesystem replacement cannot prove requested version` case.

pkg/langmgr/golang.go
omercnet: Good catch, those were silently skipped. `filterUpdatesInGoMod` now also treats modules named in the module's go.sum as applicable, and a module absent from go.mod is verified against the versions go.sum records full contents for, since those are the versions the build downloads. `/go.mod`-only entries do not count, because that code never enters the image. New cases cover a pre-1.17 go.sum-only dependency at and below the target version and the go.mod-only entry.

pkg/langmgr/golang.go
omercnet: Done. Verification tracks which updates it has proven and any failure returns `unverifiedUpdateNames`, so a workspace that aborts with later members unread reports every update still lacking proof and none of them reach the validated or VEX set. `TestUpdateGoModuleWorkspaceFailureReportsAllUnverified` drives a mid-flight read failure across two members and asserts both members' packages come back.

pkg/langmgr/golang.go
omercnet: Switched to a tagged switch on `rep.Old.Version` in `governingReplacement`. `golangci-lint run ./...` reports 0 issues on the package.

pkg/langmgr/golang.go
```diff
@@ -213,6 +216,215 @@ func buildGoUpdateCmd(modPath, allGetCmd string) string {
...
+// authoritative go.mod of its own: requirements land in the member modules, so
+// those are the modules to verify.
+func goWorkspaceMemberDirs(goWorkContent []byte, workRoot string) ([]string, error) {
+	f, err := modfile.ParseWork("go.work", goWorkContent, nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse go.work: %w", err)
+	}
+
+	var modDirs []string
+	for _, use := range f.Use {
```
robert-cronin, line 468: This only records use entries, so a go.work local replacement can still be reported as patched.

pkg/langmgr/golang.go
```diff
@@ -1012,6 +1302,28 @@ func (gm *golangManager) updateGoModule(
...
+		}
+		goSumBytes, goSumErr := gm.readOptionalGoSum(ctx, &state, target.goSumPath)
+		if goSumErr != nil {
+			return state, unverifiedUpdateNames(updates, verified), fmt.Errorf("failed to read %s after update: %w", target.goSumPath, goSumErr)
+		}
+		if verifyErr := verifyGoModUpdates(goModBytes, goSumBytes, target.updates); verifyErr != nil {
+			return state, unverifiedUpdateNames(updates, verified), fmt.Errorf("go module update verification failed for %s: %w", target.goModPath, verifyErr)
+		}
+		for _, u := range target.updates {
+			verified[u.Name] = struct{}{}
```
robert-cronin: This marks the package globally verified, so one workspace member can hide the same module failing in another.

pkg/langmgr/golang.go
```diff
@@ -1132,6 +1461,23 @@ func (gm *golangManager) upgradePackagesWithTooling(
 			llb.WithProxy(utils.GetProxy()),
 		).Root()
 
+		// Confirm the bumps survived 'go mod tidy -e' instead of trusting the
+		// exit code. Failures name the affected packages so they are reported as
+		// unpatched rather than claimed as remediated.
+		if len(moduleUpdates) > 0 {
+			goModBytes, extractErr := buildkit.ExtractFileFromState(ctx, gm.config.Client, &toolingState, "/workspace/go.mod")
+			if extractErr != nil {
+				return currentState, getPackageNames(moduleUpdates), fmt.Errorf("failed to read go.mod after update at %s: %w", modPath, extractErr)
```
robert-cronin: An early tooling-container failure still omits updates for module paths that have not been processed yet.

pkg/langmgr/golang.go
```diff
@@ -213,6 +216,215 @@ func buildGoUpdateCmd(modPath, allGetCmd string) string {
...
+		if r != nil && r.Mod.Path != "" {
+			present[r.Mod.Path] = struct{}{}
+		}
+	}
+	for _, rep := range f.Replace {
+		if rep != nil && rep.Old.Path != "" {
+			present[rep.Old.Path] = struct{}{}
+		}
+	}
+	for modPath := range goSumBuiltVersions(goSumContent) {
```
robert-cronin: Could this use go.sum only for pre-1.17 modules so stale modern checksums don't become false dependencies?

pkg/langmgr/golang.go
omercnet: Fixed in 9ac0c4b: parseGoWorkspace now returns the go.work replace directives along with the member dirs, and verification treats a workspace replacement as overriding the member's go.mod, so unversioned local contents or a different module path fail instead of being reported as patched. Covered by TestUpdateGoModuleWorkspaceReplacementIsUnprovable and the new go.work cases in TestVerifyGoModUpdates.

pkg/langmgr/golang.go
omercnet: Fixed in 9ac0c4b: the global verified set is replaced by pendingProofs, a per-package count of the modules accountable for it, so a package counts as proven only once every member that referenced it has passed. Regression test: TestUpdateGoModuleWorkspaceProofIsPerMember.

pkg/langmgr/golang.go
omercnet: Fixed in 9ac0c4b: accountability for every detected module is collected before the first module is touched, and failures now report unverifiedUpdateNames over the whole requested set, so modules queued behind the failure are never claimed as remediated. Regression test: TestUpgradePackagesWithToolingReportsUnprocessedModules.

pkg/langmgr/golang.go
omercnet: Done in 9ac0c4b: the go.sum fallback is gated on goModListsAllDependencies, so only modules whose go directive is absent or below 1.17 consult go.sum. verifyGoModUpdates still falls back to go.sum when tidy dropped a requirement, but a stale checksum there can only produce a false failure, never a false pass, so that path stays conservative. Regression test: TestFilterUpdatesInGoMod/ignores_stale_go.sum_entry_for_a_pruned_module_graph.

omercnet:
Review feedback from 2026-08-30 addressed in 9ac0c4b, with a regression test per fix:

- `go.work` replacements now travel with the member modules and override their `go.mod`, so a workspace-level redirect is no longer reported as patched (`TestUpdateGoModuleWorkspaceReplacementIsUnprovable`).
- Proof is counted per accountable module, so verifying a module in one workspace member no longer hides another member still requiring the vulnerable version (`TestUpdateGoModuleWorkspaceProofIsPerMember`).
- The tooling-container path collects module accountability before touching anything and reports every unproven package on failure, so modules queued behind the failure are not claimed as remediated (`TestUpgradePackagesWithToolingReportsUnprocessedModules`).
- The `go.sum` fallback is limited to modules whose `go` directive predates module graph pruning, so a stale checksum in a go 1.17+ module is no longer treated as a dependency (`TestFilterUpdatesInGoMod/ignores_stale_go.sum_entry_for_a_pruned_module_graph`).

Each test was confirmed to fail against the pre-fix behavior and pass after.

CI status on 9ac0c4b: 78/79 checks pass, including `Unit Test`, `lint`, CodeQL, all gateway/integration matrices, and `Test Go patching` (the job that exercises this code path).

`Test Ubuntu Chiseled secondary architectures` fails on `TestAptlessFullStatusComprehensiveFromBaselineARMv7` with `expected no-updates result`, i.e. the repatch of a freshly patched image still finds updates. Evidence that this is Ubuntu archive churn rather than a change here:

- The first attempt failed earlier in the same test with a hard mirror error: `404 Not Found` fetching `http://ports.ubuntu.com/ubuntu-ports/pool/main/a/attr/libattr1_2.5.2-1ubuntu0.1_armhf.deb`, so the armhf pool was mid-publish today.
- `TestAptlessFullStatusComprehensiveFromBaselineARM64` failed with the identical assertion on an unrelated branch today (run 33358720917, `robert-cronin/issue-1677-oci-layout-input`), which does not contain this commit.
- The same ARM64 test failed once here and then passed on rerun of the same commit, with no code change.
- This diff only touches `pkg/langmgr/golang.go`. The chiseled e2e runs `copa patch` with no `--report` and the default `--pkg-types=os`, so library updates are filtered out and `pkg/patch/core.go` never reaches `langmgr.GetLanguageManagers`.

The job was rerun three times; the ARMv7 variant keeps failing while the armhf pool settles. Happy to rerun again once the archive stabilizes.

omercnet:
Follow-up on the ARMv7 Chiseled failure: the Ubuntu ports armhf pool has settled (`libattr1_2.5.2-1ubuntu0.1_armhf.deb` now returns 200 where it returned 404 during the 10:23Z attempt, and both `2.5.2-1ubuntu0.1` and `2.5.2-1build1.1` are published). Rerunning `Test Ubuntu Chiseled secondary architectures` on the same commit passed with no code change. CI is now green: 79/79 checks on 9ac0c4b, zero unresolved review threads.

pkg/langmgr/golang.go
```diff
@@ -1064,11 +1450,38 @@ func (gm *golangManager) upgradePackagesWithTooling(
...
+
+	// Record which updates each module is accountable for before any module is
+	// touched. The same `go get` list runs against every module in the image, so
+	// verifying all of them against one module would fail a module that never
+	// depended on the vulnerable package. Collecting up front also keeps
+	// reporting honest: a failure partway through leaves the remaining modules
+	// unprocessed, and the packages they carry must still be reported as
+	// unpatched instead of assumed remediated.
+	accountableUpdates := make(map[string]unversioned.LangUpdatePackages, len(goModPaths))
+	pendingProofs := make(map[string]int, len(updates))
```
robert-cronin, line 1469: Could we reject updates that no discovered module proves? They currently return success and can enter the validated/VEX set without verification.

pkg/langmgr/golang.go
```diff
@@ -213,6 +216,271 @@ func buildGoUpdateCmd(modPath, allGetCmd string) string {
...
+	for _, u := range updates {
+		if u.FixedVersion == "" {
+			continue
+		}
+
+		requiredVersion, inGoMod := required[u.Name]
+
+		// A go.work replace overrides every member module, so it governs what
+		// the workspace build compiles no matter what the member's go.mod says.
+		if rep := governingReplacement(workReplaces, u.Name, requiredVersion); rep != nil {
```
robert-cronin, line 265: Could we resolve version-qualified go.work replacements before falling back to go.sum? A pre-1.17 indirect dependency can pass verification while the workspace still builds local code.

PR #1685, opened 2026-08-26
fix(test): use moby container and network types in bulk patch test

omercnet:
testcontainers-go 0.44.0 (#1568) moved `HostConfigModifier` and `PortBindings` to the `github.com/moby/moby/api` types, so `integration/bulk/singlearch_patch_test.go` stopped compiling. That breaks `Run golangci-lint` (typecheck) and `Test Bulk Patching` on main itself, and consequently on every open pull request.

This ports the registry helper to `network.PortMap` with a `netip.Addr` host IP and retidies `github.com/docker/go-connections` to indirect, since that was its last direct use.

Verified locally with `go vet ./integration/...`, `go build ./...`, `go mod tidy -diff` clean, and gofumpt clean.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -18,7 +18,6 @@ require (
 	github.com/docker/buildx v0.32.1
 	github.com/docker/cli v29.5.3+incompatible
 	github.com/docker/docker v28.5.2+incompatible
-	github.com/docker/go-connections v0.7.0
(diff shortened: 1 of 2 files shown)


robert-cronin:
This exact patch has already landed on main via #1683 (2a7738b), so I think this PR can be closed as redundant.

PR #1696, opened 2026-09-02
test: make LogPipe test deterministic

robert-cronin:
## Summary

- replace the timing-based `TestLogPipe` assertion with deterministic `io.Pipe` completion synchronization
- avoid concurrent reads and writes to the shared output buffer
- restore the global Logrus output after the test

## Testing

- `go test ./pkg/utils -run '^TestLogPipe$' -count=1000`
- `go test -race ./pkg/utils -run '^TestLogPipe$' -count=100`
- `make test`
- `make lint`

### DIFF
--- a/pkg/utils/log_test.go
+++ b/pkg/utils/log_test.go
@@ -3,7 +3,7 @@ package utils
 import (
 	"bytes"
 	"fmt"
-	"os/exec"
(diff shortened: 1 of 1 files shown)


ashnamehrotra (approved):
LGTM

PR #1711, opened 2026-09-10
fix: attach registry credentials to the OCI layout export solve

mjameeghouri:
Add registry credentials to the OCI layout export solve, so an image whose source lives in a private registry can be patched.

## Problem

`copa patch` against a source in a private registry patches every platform successfully and then fails in the export:

```
✓ linux/amd64      Patched      Successfully patched
✓ linux/arm64      Patched      Successfully patched
Creating OCI layout from 2 patched platforms only
Creating OCI layout from 2 BuildKit states
failed to create OCI layout: failed to solve platform: failed to solve:
failed to load cache key: failed to authorize:
failed to fetch anonymous token: unexpected status from GET request to
https://<registry>/v2/token?scope=repository%3A<repo>%3Apull&service=<registry>: 403 Forbidden
```

The patched states still reference the source image, so BuildKit resolves the `docker-image://` source again while solving the export. The patch solve attaches the caller's Docker credentials through `pkg/patch.authenticatedSolveOpt`, but the export solves in `pkg/buildkit/buildkit.go` construct their `client.SolveOpt` with only `Exports` set — that package does not import `session/auth/authprovider` at all — so BuildKit resolves the source anonymously.

An anonymous read succeeds against a public source and fails against any private one. The practical effect is that an image already sitting in a private registry cannot be patched, which includes re-patching an image copa itself produced there earlier.

`remote.WithAuthFromKeychain` is already used elsewhere in the package, but it cannot help here: the failing read is a BuildKit solve rather than a go-containerregistry call, so it needs a session attachable, not a keychain.

## Changes

- Add `ensureAuthSession`, which attaches the same Docker credentials the patch solve uses, unless the caller supplied a session of its own.
- Call it from `solvePlatformOCI`. All three export paths — single-platform, multi-platform, and the mixed patched/preserved layout — solve through that function, so a single place covers them.

Loading credentials through `authprovider.LoadAuthConfig(config.LoadDefaultConfigFile(...))` means configured credential helpers work, not just static `auths` entries — which matters for registries whose credentials come from a helper binary, such as Google Artifact Registry via `gcloud auth configure-docker`.

## Testing

- `go build ./...`, `go vet ./pkg/buildkit/`, and `go test ./pkg/buildkit/...` all pass.
- Added `TestEnsureAuthSessionAttachesDockerCredentials`, covering: a session is attached when none is set, a caller-supplied session is left untouched, and a nil `SolveOpt` does not panic. Confirmed the first case fails if the helper is made a no-op, so the test genuinely covers the fix rather than passing vacuously.
- Reproduced the original failure against a private Google Artifact Registry source on v0.14.2 and again on v0.15.0. `crane manifest` against the same reference succeeds, since go-containerregistry runs the configured credential helper — which is what isolated the failure to the BuildKit solve.

## Notes

[#1419](https://github.com/project-copacetic/copacetic/pull/1419) described this same symptom and named `exportOriginalImagePlatformsAsOCI()`, but was closed unmerged and approached it via the ggcr keychain. This takes the session route instead, at the one function all export solves pass through.


robert-cronin (approved):
lgtm, thank you! CI failures appear to be pre-existing on main and unrelated

